Issue #9, opened 2020-09-04
Signal color space in image formats using ICC Profiles


simontWork:
I think most are in agreement that magic strings and invaild ICC profiles are not a good way of handling still image formats.

Could we explore:

1. Creating ICCMax profiles (with the help of ICC), or
2. using a floating point linear image format, such as that described in ITU-R BT.2100?

palemieux:
What is the context of this issue?

simontWork:
Still image formats are discussed in the document, but no standards compliant method is outlined.

palemieux:
> but no standards compliant method is outlined.

What method(s) do you have in mind?

* signaling the color space information
* carrying the color space information and the image pixels
* encoding the image pixels

digitaltvguy:
This issue is being discussed in ICC HDR Working Group.  Is there a liason to ICC in W3C?

ProgramMax:
FWIW Chrome already supports ICC profiles in most image formats.
When you say "still image formats" I assume this is as opposed to video? But I think animated images might also belong in this group.

I would like to add a restriction to ICC profile use in file formats: the ICC profile must come before the frame data. Mirroring what I said in #11, web browsers display the image as it is being downloaded. If the image is fully displayed before the browser learns the image should be in a different color space, the user will experience a visual pop.

For animated images like GIF, APNG, and animated WebP, I might need the wisdom of the entertainment folks. Would it be safe to allow only one ICC profile for a whole image? Would a movie have one scene in color space A and another in color space B? Or a TV channel where one show is color space A and the next is color space B?

Once I learn more about that, I can provide some insight into what would happen if the ICC profile comes between frames of an animated image.

Chrome currently does not support ICC profiles in GIFs because all of this is under-specified. There also doesn't seem to be an established "street spec" that everyone agrees upon. I previously contacted the US Library of Congress to recommend they reconsider ICC profile support in GIF as one of its benefits.

svgeesus:
> I think most are in agreement that magic strings and invaild ICC profiles are not a good way of handling still image formats.

Yes. (The context is [Using the ITU BT.2100 PQ EOTF with the PNG Format](https://www.w3.org/TR/png-hdr-pq/) which embeds an invalid ICC profile, basically as a flag).

svgeesus:
> This issue is being discussed in ICC HDR Working Group. 

It is, yes.

> Is there a liaison to ICC in W3C?

[...]

PR #12, opened 2021-01-26
Add support for HDR imagery to the PNG format

### LINKED ISSUE #11: Create new CICP chunk in PNG (See ITU-T H.273)
Can we add a chunk to the PNG standard which stores "Coding-independent code points for video signals" (CICP) so that graphics content identification can be identified the same way motion video images are?

This would prevent an issue where an ICC profile gets stripped from a file causing a loss of HDR/WCG signaling.

Color Primaries, Transfer Function and Matrix Coefficients are identified in Tables 2, 3 & 4 of ITU-T H.273.

We would also need to store signal range.  This requires 7 bytes according to Pierre.

ProgramMax:
This commit adds a proposal to ammend the PNG specification to
add support for a new cICP chunk that enabled HDR support.

Closes #11

### DIFF
--- a/png_cicp_chunk_proposal.md
+++ b/png_cicp_chunk_proposal.md
@@ -6,19 +6,7 @@ Status: Draft
 ## Problem to be solved
 The gAMA chunk of the Portable Network Graphics (PNG) format (specified in [PNG]) parameterized the transfer function of the image as a power law. As such, it cannot model the Reference PQ or HLG OOTFs specified in [BT2100-1], which are commonly used for HDR images.
 
-This specification uses the existing iCCP chunk to unambiguously signal the color system of an image that uses the Reference PQ EOTF specified in [BT2100-1]. It also allows graceful processing by decoders that do not conform to this specification by recommending fallback values for the gAMA chunk, cHRM chunk, and embedded ICC profile.
-
 An existing W3C group note (https://www.w3.org/TR/png-hdr-pq/) specifies an approach which is limited: it supports signaling of only the ITU BT.2100 PQ EOTF and uses magic values in the iCCP chunk to signal color spaces.
-
-[H.273](https://www.itu.int/rec/T-REC-H.273/en) specifies a controlled vocabulary for the parameterization of
-color space information:
-* Color Primaries
-* Transfer Function
-* Matrix Coefficients (only required if essence is stored as Y’CbCr)
-
-Use [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) “recommendations” to signal the controlled vocabulary for image formats commonly used for baseband linear broadcasts or file-based Video-on-Demand(VOD) services.
-* Tables within this document summarize values most commonly used in broadcast for 47 different standards documents including H.273
-
 
 ## Basic requirements
 * does not break current implementations
@@ -37,6 +25,17 @@ H.273 color space parameters:
 
 [ed.: these are inspired from recent JPEG standards that incorporate
 H.273 color space parameters]
+
+This specification uses the existing iCCP chunk to unambiguously signal the color system of an image that uses the Reference PQ EOTF specified in [BT2100-1]. It also allows graceful processing by decoders that do not conform to this specification by recommending fallback values for the gAMA chunk, cHRM chunk, and embedded ICC profile.
+
+[H.273](https://www.itu.int/rec/T-REC-H.273/en) specifies a controlled vocabulary for the parameterization of
+color space information:
+* Color Primaries
+* Transfer Function
+* Matrix Coefficients (only required if essence is stored as Y’CbCr)
+
+Use [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) “recommendations” to signal the controlled vocabulary for image formats commonly used for baseband linear broadcasts or file-based Video-on-Demand(VOD) services.
+* Tables within this document summarize values most commonly used in broadcast for 47 different standards documents including H.273
 
 ## A. References
 ### A.1 Normative references
--- a/png_cicp_chunk_proposal.md
+++ b/png_cicp_chunk_proposal.md
@@ -20,11 +20,12 @@ H.273 color space parameters:
 
 * COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
 * TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
-* MATCOEFFS, 2 bytes, One of the MatrixCoefficients enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
 * VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
 
 [ed.: these are inspired from recent JPEG standards that incorporate
 H.273 color space parameters]
+
+[ed.: The MATOEFFS parameter is not included because PNG does not support YCbCr]
 
 This specification uses the existing iCCP chunk to unambiguously signal the color system of an image that uses the Reference PQ EOTF specified in [BT2100-1]. It also allows graceful processing by decoders that do not conform to this specification by recommending fallback values for the gAMA chunk, cHRM chunk, and embedded ICC profile.
 
--- a/png_cicp_chunk_proposal.md
+++ b/png_cicp_chunk_proposal.md
@@ -40,6 +40,8 @@ When the cICP chunk is present, PNG decoders that recognize it shall ignore the 
 NOTE: [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) summarize combinations of H.273 parameters corresponding to common baseband linear broadcasts and file-based Video-on-Demand(VOD) services.
 
 ## A. References
+
+[ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I). Series H: Audiovisual and multimedia systems - Usage of video signal type code points
 [BT2100-1]
 [Recommendation ITU-R BT.2100-1](https://www.itu.int/rec/R-REC-BT.2100), Image parameter values for high dynamic range television for use in production and international programme exchange
 
--- a/png_cicp_chunk_proposal.md
+++ b/png_cicp_chunk_proposal.md
@@ -18,9 +18,9 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
 Define a cICP chunk that contains the 7 bytes necessary to carry the
 H.273 color space parameters:
 
-* COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23091-2
-* TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23091-2
-* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | ISO/IEC 23091-2
+* COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
 
 [ed.: these are inspired from recent JPEG standards that incorporate
 H.273 color space parameters]
@@ -51,3 +51,6 @@ Use [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-
 
 [BT2100-in-PNG]
 [Using the ITU BT.2100 PQ EOTF with the PNG format](https://www.w3.org/TR/png-hdr-pq/)
+
+[ISO/IEC 23091-2]
+[ISO/IEC 23091-2](https://www.iso.org/standard/81546.html)
--- a/png_cicp_chunk_proposal.md
+++ b/png_cicp_chunk_proposal.md
@@ -6,7 +6,7 @@ Status: Draft
 ## Problem to be solved
 The gAMA chunk of the Portable Network Graphics (PNG) format (specified in [PNG]) parameterized the transfer function of the image as a power law. As such, it cannot model the Reference PQ or HLG OOTFs specified in [BT2100-1], which are commonly used for HDR images.
 
-An existing W3C group note (https://www.w3.org/TR/png-hdr-pq/) specifies an approach which is limited: it supports signaling of only the ITU BT.2100 PQ EOTF and uses magic values in the iCCP chunk to signal color spaces.
+An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limited: it supports signaling of only the ITU BT.2100 PQ EOTF and uses magic values in the iCCP chunk to signal color spaces.
 
 ## Basic requirements
 * does not break current implementations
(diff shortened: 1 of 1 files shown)


png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
...
+* Matrix Coefficients (only required if essence is stored as Y’CbCr)
+
+Use [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) “recommendations” to signal the controlled vocabulary for image formats commonly used for baseband linear broadcasts or file-based Video-on-Demand(VOD) services.
+* Tables within this document summarize values most commonly used in broadcast for 47 different standards documents including H.273
+
+
+## Basic requirements
+* does not break current implementations
+* extensible signaling of color space based on H.273
+* does not require the presence of iCCP chunk and embedded ICC profiles
```
lrosenthol, line 26: Why is this a requirement?

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
...
+* cICP chunk comes before frame data
```
lrosenthol, line 27: Do you mean before the `IDAT` chunk?   Can you clarify this in terms of which chunk can come before and/or after (as per PNG spec, 5.6)

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
...
+* Transfer Function
+* Matrix Coefficients (only required if essence is stored as Y’CbCr)
+
+Use [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) “recommendations” to signal the controlled vocabulary for image formats commonly used for baseband linear broadcasts or file-based Video-on-Demand(VOD) services.
+* Tables within this document summarize values most commonly used in broadcast for 47 different standards documents including H.273
+
+
+## Basic requirements
+* does not break current implementations
+* extensible signaling of color space based on H.273
```
lrosenthol, line 25: Why only H.273?   Why not support other forms of HDR?

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
...
+## Strawman approach
+Define a cICP chunk that contains the 7 bytes necessary to carry the
+H.273 color space parameters:
+
+* COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
+* TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
+* MATCOEFFS, 2 bytes, One of the MatrixCoefficients enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
+* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
+
+[ed.: these are inspired from recent JPEG standards that incorporate
```
lrosenthol, line 38: JPEG-XL uses Rec. ITU-T H.273 | ISO/IEC 23091-2:2019(E) as the normative reference - why are you using 23001-8?

lrosenthol: I also don't see how these four values map into the values described in JPEG-XL A.4 (Color Encoding).

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Basic requirements
+* does not break current implementations
+* extensible signaling of color space based on H.273
+* does not require the presence of iCCP chunk and embedded ICC profiles
+* cICP chunk comes before frame data
+
+## Strawman approach
+Define a cICP chunk that contains the 7 bytes necessary to carry the
```
lrosenthol, line 30: `cICP` would not be a valid chunk name (as per PNG spec, 5.4), specifically the four character since I would consider it "safe to copy" which means it should be lowercase - thus `cICp`

lrosenthol (changes requested):
Raising a bunch of questions and concerns

png_cicp_chunk_proposal.md
svgeesus: I agree with @lrosenthol that "before IDAT" is the best way to express this.

png_cicp_chunk_proposal.md
svgeesus: It is not clear to me whether this is intended to be safe to copy or not. But  iCCP  is also not safe to copy.

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
...
+H.273 color space parameters:
+
+* COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
+* TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
+* MATCOEFFS, 2 bytes, One of the MatrixCoefficients enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
+* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | ISO/IEC 23001-8
+
+[ed.: these are inspired from recent JPEG standards that incorporate
+H.273 color space parameters]
+
```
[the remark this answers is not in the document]
svgeesus: Yes, priority in the presence of conflicting information is essential to specify.

As an example, look at what the PNG spec [says about the `iCCP` chunk in the presence of `gAMA` and cHRM` for readers and writers](https://www.w3.org/TR/PNG/#11iCCP).

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
+# Adding support for HDR imagery to the PNG format
+Authors (alphabetical): Chris Blume, Chris Seeger, Pierre-Anthony Lemieux
+
+Status: Draft
+
+## Problem to be solved
+The gAMA chunk of the Portable Network Graphics (PNG) format (specified in [PNG]) parameterized the transfer function of the image as a power law. As such, it cannot model the Reference PQ or HLG OTFs specified in [BT2100-1], which are commonly used for HDR images.
+
+This specification uses the existing iCCP chunk to unambiguously signal the color system of an image that uses the Reference PQ EOTF specified in [BT2100-1]. It also allows graceful processing by decoders that do not conform to this specification by recommending fallback values for the gAMA chunk, cHRM chunk, and embedded ICC profile.
```
svgeesus, line 9: "This specification" which specification?

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
...
+
+An existing W3C group note (https://www.w3.org/TR/png-hdr-pq/) specifies an approach which is limited: it supports signaling of only the ITU BT.2100 PQ EOTF and uses magic values in the iCCP chunk to signal color spaces.
```
svgeesus, line 11: I guess "magic values" is sufficient to indicate that this is a really bad idea. Specifically, the ICC profile, if applied, does not give the correct rendering and must be ignored :) and only its presence is used as a verbose way to label the content. This proposal is thus a significant improvement on that, in terms of internal consistency and also filesize.

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,50 @@
...
+
+[H.273](https://www.itu.int/rec/T-REC-H.273/en) specifies a controlled vocabulary for the parameterization of
+color space information:
+* Color Primaries
+* Transfer Function
+* Matrix Coefficients (only required if essence is stored as Y’CbCr)
```
svgeesus, line 17: So, PNG does not support Y'CbCr; only RGB, RGBA, G, and GA.
Why is this Matrix Coefficients field needed?

png_cicp_chunk_proposal.md
svgeesus, line 11: Also, add that note to the informative references

png_cicp_chunk_proposal.md
svgeesus: Filesize is one reason, a chunk with a 7 byte payload is smaller than an ICC profile with zlib compression.

png_cicp_chunk_proposal.md
svgeesus: Nor indeed is `gAMA` and `cHRM`

svgeesus (approved):
Approving because beyond a certain point, it is clearer to merge the pull request and then make additional comments against a directly viewable document.

png_cicp_chunk_proposal.md
ProgramMax: Agreed. I will update it.

png_cicp_chunk_proposal.md
lrosenthol: True, but it doesn't give rich fallback then.  Also doesn't align with the work going on at the ICC to consider a `CICP` tag.

png_cicp_chunk_proposal.md
lrosenthol: Why would you consider the profile and gamma info not safe to copy?  They are absolutely safe to copy since the assumption is that you aren't changing the colorspace of the image data when processing it.

png_cicp_chunk_proposal.md
ProgramMax: I'm not sure I follow.
You are suggesting it should require the other chunk? Would you mind helping me understand why it should be required to have both?

Would it be an acceptable fallback to allow both chunks but not require both chunks? (If so, we probably want to explicitly specify how to handle a file with both chuns.)

png_cicp_chunk_proposal.md
ProgramMax: I don't know enough here to have answer.
Would you mind elaborating?

png_cicp_chunk_proposal.md
ProgramMax: Looking at http://w3.org/TR/PNG-Structure.html#Chunk-naming-conventions I think I understand why those chunks are unsafe to copy and why this new chunk should be as well.

"If the program has made _any_ changs to _critical_ chunks, ... then unrecognized unsafe chunks must **not** be copied to the output PNG file. (Of course, if the program **does** recognize the chunk, it can choose to output an appropriately modified version.)"

So for example, if the program doesn't understand the HDR chunk and modifies some pixel values, it might do so in the wrong color space. I don't think we want to preserve the incorrect color space. I think we want to throw out the HDR information and now have a file in sRGB, for example.

png_cicp_chunk_proposal.md
palemieux: @lrosenthol [H.273](https://www.itu.int/rec/T-REC-H.273/en) supports common SDR/HDR formats. Do you know of any HDR format that PNG should support but is not listed in H.273?

png_cicp_chunk_proposal.md
palemieux: Per the 2021-01-12 meeting:

>     - two challenges re: signaling colorspace in the ICC profile in contrast to signaling colorspace directly in the image:
>         - adds a level of indirection
>         - ICC profile can/is be easily stripped

I think HDR-in-PNG signaling should not be contingent on ICC profiles, although ICC profiles can be included, e.g. for fallback.

png_cicp_chunk_proposal.md
palemieux: Rec. ITU-T H.273 was developed as twin-text with ISO/IEC 23001-8, but it looks like ISO has replaced ISO/IEC 23001-8 with ISO/IEC 23091-2.

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,53 @@
...
+* COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23091-2
+* TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | ISO/IEC 23091-2
+* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | ISO/IEC 23091-2
+
+[ed.: these are inspired from recent JPEG standards that incorporate
+H.273 color space parameters]
+
+[ed.: The MATOEFFS parameter is not included because PNG does not support YCbCr]
+
+This specification uses the existing iCCP chunk to unambiguously signal the color system of an image that uses the Reference PQ EOTF specified in [BT2100-1]. It also allows graceful processing by decoders that do not conform to this specification by recommending fallback values for the gAMA chunk, cHRM chunk, and embedded ICC profile.
```
palemieux, line 30:
```suggestion
When the cICP chunk is present, PNG decoders that recognize it shall ignore the following chunks:
- iCCP
- gAMA 
- cHRM 
- sRGB 
```

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,53 @@
...
+color space information:
+* Color Primaries
+* Transfer Function
+* Matrix Coefficients (only required if essence is stored as Y’CbCr)
+
+Use [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) “recommendations” to signal the controlled vocabulary for image formats commonly used for baseband linear broadcasts or file-based Video-on-Demand(VOD) services.
+* Tables within this document summarize values most commonly used in broadcast for 47 different standards documents including H.273
+
+## A. References
+### A.1 Normative references
```
palemieux, line 42:
```suggestion
```

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,57 @@
...
+
+The cICP chunk comes before IDAT chunk.
+
+When the cICP chunk is present, PNG decoders that recognize it shall ignore the following chunks:
+- iCCP
+- gAMA 
+- cHRM 
+- sRGB 
+
+Use [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) “recommendations” to signal the controlled vocabulary for image formats commonly used for baseband linear broadcasts or file-based Video-on-Demand(VOD) services.
```
palemieux, line 40:
```suggestion
NOTE: [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) summarize combinations of H.273 parameters corresponding to common baseband linear broadcasts and file-based Video-on-Demand(VOD) services.
```

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,57 @@
...
+* Tables within this document summarize values most commonly used in broadcast for 47 different standards documents including H.273
```
palemieux, line 41:
```suggestion
```

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,57 @@
...
+
+## A. References
```
palemieux, line 42:
```suggestion
## A. References

[ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I). Series H: Audiovisual and multimedia systems - Usage of video signal type code points
```

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,58 @@
+# Adding support for HDR imagery to the PNG format
+Authors (alphabetical): Chris Blume, Pierre-Anthony Lemieux, Chris Seeger
```
palemieux, line 2:
```suggestion
Editors: Chris Blume, Pierre-Anthony Lemieux, Chris Seeger
```

png_cicp_chunk_proposal.md
lrosenthol: But now that means that PNG has its own model for colorspace specification that differs from other formats (eg. JPEG)

lrosenthol: Also, what if the embedded ICC profile *has* the new tag for HDR - why not support that?

png_cicp_chunk_proposal.md
```diff
@@ -0,0 +1,58 @@
...
+
+[ed.: these are inspired from recent JPEG standards that incorporate
+H.273 color space parameters]
+
+[ed.: The MATOEFFS parameter is not included because PNG does not support YCbCr]
+
+The cICP chunk comes before IDAT chunk.
+
+When the cICP chunk is present, PNG decoders that recognize it shall ignore the following chunks:
+- iCCP
```
lrosenthol, line 35: See above - this is incorrect in the presence of an ICC profile containing an CICP tag.

lrosenthol (changes requested):
Still not aligned with the rest of the imaging world...

png_cicp_chunk_proposal.md
svgeesus: > 
> 
> But now that means that PNG has its own model for colorspace specification that differs from other formats (eg. JPEG)

PNG also has its own PNG-specific way to signal sRGB, that doesn't use an ICC profile.
EXIF has an EXIF-specific way to signal either sRGB or Adobe 1998 RGB, without an ICC profile

svgeesus (approved):
Decided to merge at 9 Feb 2021 ColorWeb CG mtg

lrosenthol:
I agree to submit a PR that connects this to #9 and the problems with iCCP vs. CICP

PR #13, opened 2021-02-09
Wrote up proposal around both tags

lrosenthol:
This proposal attempts to reconcile some of the items in #9 and the new tag:
- File format restriction
- Processor handling

Also did some basic editorial cleanup.

### DIFF
--- a/png_cicp_chunk_proposal.md
+++ b/png_cicp_chunk_proposal.md
@@ -13,31 +13,33 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
 * extensible signaling of color space based on H.273
 * does not require the presence of iCCP chunk and embedded ICC profiles
 
+## Non Requirements
+* compatibility with HDR colorspace definitions in any other raster image format (eg. JPEG, AVIF)
+
 ## Strawman approach
-[H.273](https://www.itu.int/rec/T-REC-H.273/en) specifies a controlled vocabulary for the parameterization of
-color space information.
+[H.273](https://www.itu.int/rec/T-REC-H.273/en) specifies a controlled vocabulary for the parameterization of color space information.
 
-Define a cICP chunk that contains the 7 bytes necessary to carry the
-H.273 color space parameters:
+Define a `cICP` chunk that contains the 7 bytes necessary to carry the H.273 color space parameters:
 
-* COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
-* TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
-* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **COLPRIMS**, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **TRANSFC**, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **VIDFRNG**, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
 
-[ed.: these are inspired from recent JPEG standards that incorporate
-H.273 color space parameters]
+[ed.: these are inspired from recent JPEG standards (eg. JPEG-XL) that incorporate H.273 color space parameters]
 
 [ed.: The MATOEFFS parameter is not included because PNG does not support YCbCr]
 
-The cICP chunk comes before IDAT chunk.
-
-When the cICP chunk is present, PNG decoders that recognize it shall ignore the following chunks:
-- iCCP
-- gAMA 
-- cHRM 
-- sRGB 
-
 NOTE: [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) summarize combinations of H.273 parameters corresponding to common baseband linear broadcasts and file-based Video-on-Demand(VOD) services.
+
+### Implementation Details
+The `cICP` chunk SHALL come before `IDAT` chunk.  
+
+A PNG SHALL not contain both a `cICP` chunk and an `iCCP` chunk. If a PNG decoder detects the presence of both a `cICP` and a `iCCP` chunk, the behavior is undefined.
+
+When the `cICP` chunk is present a PNG decoder SHALL ignore the following chunks:
+- `gAMA`
+- `cHRM` 
+- `sRGB` 
 
 ## A. References
 
--- a/png_cicp_chunk_proposal.md
+++ b/png_cicp_chunk_proposal.md
@@ -1,5 +1,5 @@
 # Adding support for HDR imagery to the PNG format
-Editors: Chris Blume, Pierre-Anthony Lemieux, Chris Seeger
+Editors: Chris Blume, Pierre-Anthony Lemieux, Chris Seeger, Leonard Rosenthol
 
 Status: Draft
 
@@ -14,29 +14,28 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
 * does not require the presence of iCCP chunk and embedded ICC profiles
 
 ## Non Requirements
-* compatibility with HDR colorspace definitions in any other raster image format (eg. JPEG, AVIF)
+* bit-identical serialization of the H.273 color space parameters as used by other raster image formats (eg. JPEG, AVIF)
 
 ## Strawman approach
 [H.273](https://www.itu.int/rec/T-REC-H.273/en) specifies a controlled vocabulary for the parameterization of color space information.
 
-Define a `cICP` chunk that contains the 7 bytes necessary to carry the H.273 color space parameters:
+Define a `cICP` chunk that contains the 3 bytes necessary to carry the H.273 color space parameters:
 
-* **COLPRIMS**, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
-* **TRANSFC**, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **COLPRIMS**, 1 byte, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **TRANSFC**, 1 byte, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
 * **VIDFRNG**, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
 
-[ed.: these are inspired from recent JPEG standards (eg. JPEG-XL) that incorporate H.273 color space parameters]
+NOTE: While these are inspired from recent JPEG standards (eg. JPEG-XL) that incorporate these color space parameters, this specification is only using 1 byte per value as defined in H.273 (vs. 2 bytes in JPEG-XL).
 
 [ed.: The MATOEFFS parameter is not included because PNG does not support YCbCr]
 
 NOTE: [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) summarize combinations of H.273 parameters corresponding to common baseband linear broadcasts and file-based Video-on-Demand(VOD) services.
 
-### Implementation Details
 The `cICP` chunk SHALL come before `IDAT` chunk.  
 
-A PNG SHALL not contain both a `cICP` chunk and an `iCCP` chunk. If a PNG decoder detects the presence of both a `cICP` and a `iCCP` chunk, the behavior is undefined.
+A PNG MAY contain both a `cICP` chunk and an `iCCP` chunk. If a PNG decoder detects the presence of both a `cICP` and a `iCCP` chunk, the behavior is undefined.
 
-When the `cICP` chunk is present a PNG decoder SHALL ignore the following chunks:
+When the `cICP` chunk is present, a PNG decoder SHALL ignore the following chunks:
 - `gAMA`
 - `cHRM` 
 - `sRGB` 


png_cicp_chunk_proposal.md
```diff
@@ -13,31 +13,33 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
...
-Define a cICP chunk that contains the 7 bytes necessary to carry the
-H.273 color space parameters:
+Define a `cICP` chunk that contains the 7 bytes necessary to carry the H.273 color space parameters:
 
-* COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
-* TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
-* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **COLPRIMS**, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **TRANSFC**, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **VIDFRNG**, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
```
wantehchang, line 26: VideoFullRangeFlag is used in conjunction with MatrixCoefficients in ITU-T H.273, so VIDFRNG does not seem necessary. Is it for limited-range RGB? Thanks.

png_cicp_chunk_proposal.md
```diff
@@ -13,31 +13,33 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
...
-[ed.: these are inspired from recent JPEG standards that incorporate
-H.273 color space parameters]
+[ed.: these are inspired from recent JPEG standards (eg. JPEG-XL) that incorporate H.273 color space parameters]
 
 [ed.: The MATOEFFS parameter is not included because PNG does not support YCbCr]
 
-The cICP chunk comes before IDAT chunk.
+NOTE: [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) summarize combinations of H.273 parameters corresponding to common baseband linear broadcasts and file-based Video-on-Demand(VOD) services.
+
+### Implementation Details
```
palemieux, line 34: Why is this an implementation detail? Isn't part of the strawman?

png_cicp_chunk_proposal.md
```diff
@@ -13,31 +13,33 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
 * extensible signaling of color space based on H.273
 * does not require the presence of iCCP chunk and embedded ICC profiles
 
+## Non Requirements
+* compatibility with HDR colorspace definitions in any other raster image format (eg. JPEG, AVIF)
```
palemieux, line 17: I am not sure why this non-requirement is necessary? Can you provide an example?

png_cicp_chunk_proposal.md
palemieux: > so VIDFRNG does not seem necessary. 

I think you are right as long as we do not need to carry narrow-range RGB.

@digitaltvguy Any requirement for narrow-range RGB?

palemieux (changes requested):

palemieux:
@lrosenthol Is this intended to close #9 ?

png_cicp_chunk_proposal.md
simontWork: If you're storing a video frame that has details in the sub-blacks or super-whites (e.g. the output from a live camera), you may want to store it as narrow range, otherwise there will be clipping which will affect downstream filters.

png_cicp_chunk_proposal.md
```diff
@@ -13,31 +13,33 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
...
 
-When the cICP chunk is present, PNG decoders that recognize it shall ignore the following chunks:
-- iCCP
-- gAMA 
-- cHRM 
-- sRGB 
+A PNG SHALL not contain both a `cICP` chunk and an `iCCP` chunk. If a PNG decoder detects the presence of both a `cICP` and a `iCCP` chunk, the behavior is undefined.
 
-NOTE: [ITU-T Series H Supplement 19](https://www.itu.int/rec/T-REC-H.Sup19-201910-I) summarize combinations of H.273 parameters corresponding to common baseband linear broadcasts and file-based Video-on-Demand(VOD) services.
+When the `cICP` chunk is present a PNG decoder SHALL ignore the following chunks:
```
simontWork, line 39: Agreed, but it would be could to keep the gamma and chrominance tags, just so a legacy PNG decoder can display an image.

png_cicp_chunk_proposal.md
lrosenthol: I wanted to separate out some of the background from the details...

png_cicp_chunk_proposal.md
lrosenthol: It's necessary because it explains why we aren't "just using ICC profiles" (which would be the more compatible method).

png_cicp_chunk_proposal.md
lrosenthol: This isn't precluding them in the file - it is just saying that they are ignored during decoding.

png_cicp_chunk_proposal.md
```diff
@@ -13,31 +13,33 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
...
 
-Define a cICP chunk that contains the 7 bytes necessary to carry the
-H.273 color space parameters:
+Define a `cICP` chunk that contains the 7 bytes necessary to carry the H.273 color space parameters:
 
-* COLPRIMS, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
-* TRANSFC, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
-* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **COLPRIMS**, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **TRANSFC**, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
```
wantehchang, line 25: Although these two fields are two bytes in the ColourInformationBox (the 'colr' box) in ISO BMFFL

```
    unsigned int(16) colour_primaries;
    unsigned int(16) transfer_characteristics;
```

ITU-T H.273 says:

> An 8-bit field should be adequate for representation of the ColourPrimaries code point.

and

> An 8-bit field should be adequate for representation of the TransferCharacteristics code point

And indeed both HEVC (H.265) and AV1 represent these with 8-bit syntax elements. See page 415 in Rec. ITU-T H.265 v7 (11/2019) and page 32 in _AV1 Bitstream & Decoding Process Specification, Version 1.0.0 with Errata 1_. (They are also 8-bit syntax elements in H.274, used by VVC.)

So we may want to use only one byte each for COLPRIMS and TRANSFC.

Also, VIDFRNG can be represented with one bit.

png_cicp_chunk_proposal.md
palemieux: @lrosenthol I do not think it is a fair assessment that ICC profiles are the only way to do HDR in other raster formats. For example, they are not used in common post-production formats like MXF. Even recent versions of J2K supports both ICC profiles and H.273 signaling. I therefore do not agree that compatibility with other formats is a non-requirement.

png_cicp_chunk_proposal.md
palemieux: PNG was never intended for narrow range... so I am not convinced it is a good idea to start now.

png_cicp_chunk_proposal.md
svgeesus: I agree with @palemieux that ICC profiles are not the only way.

The specific ones cited are particularly unfortunate; JPEG is a compression scheme, not an image format and AVIF is an image format, which mainly uses H.273 signalling (ICC profiles are also supported, but not widely used, as AVIF is basically a single-frame or few-frames video).

png_cicp_chunk_proposal.md
svgeesus: I have not seen narrow-range data stored in PNG images.

On the other hand, single frames from narrow-range video would be easier to store in PNG if a narrow-to-full rescaling was not required.

png_cicp_chunk_proposal.md
lrosenthol: I accept that ICC profiles wouldn't necessarily be the only consistent way across formats...but right now they are.  And my point is that we aren't concerned with consistency in this proposal

lrosenthol: So I'll change compatibility to consistency

lrosenthol:
> @lrosenthol Is this intended to close #9 ?

No,  it it helps...

png_cicp_chunk_proposal.md
LarsBrg: Please include VideoFullRangeFlag 
VideoFullRangeFlag can also be used for RGB images, so it is relevant for some use case.
There are several narrow range RGB files in use as most apps don't support YCC files.

LarsBrg: Why are the fields 2 bytes? H.273 supports only 1 byte.

png_cicp_chunk_proposal.md
```diff
@@ -13,31 +13,33 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
...
+
+### Implementation Details
+The `cICP` chunk SHALL come before `IDAT` chunk.  
 
-When the cICP chunk is present, PNG decoders that recognize it shall ignore the following chunks:
-- iCCP
-- gAMA 
-- cHRM 
-- sRGB 
+A PNG SHALL not contain both a `cICP` chunk and an `iCCP` chunk. If a PNG decoder detects the presence of both a `cICP` and a `iCCP` chunk, the behavior is undefined.
```
ProgramMax, line 37: I don't think we should add this.

From the current PNG spec on the iCCP chunk [1]: "A PNG encoder that writes the iCCP chunk is encouraged to also write gAMA and cHRM chunks that approximate the ICC profile, to provide compatibility with applications that do not use the iCCP chunk. When the iCCP chunk is present, PNG decoders that recognize it and are capable of colour management [ICC] shall ignore the gAMA and cHRM chunks and use the iCCP chunk instead and interpret it according to [ICC-1] and [ICC-1A]. PNG decoders that are used in an environment that is incapable of full-fledged colour management should use the gAMA and cHRM chunks if present."

If a PNG cannot contain both chunks, encoders are forced to pick one without knowing what the decoder supports. Suppose my encoder writes the cICP chunk but the decoder only understands iCCP. Now I get neither.

Instead, we could follow that existing iCCP chunk specification and allow one of the chunks to be a fallback. And that is how it was already written.

[1] https://www.w3.org/TR/2003/REC-PNG-20031110/#11iCCP

png_cicp_chunk_proposal.md
ProgramMax: I think we should not have this callout.

As I explain below, we should allow both cICP and iCCP chunks. Compatibility & consistency remains if we allow both.
But even if we didn't, using a standard like H.273 is quite explicitly a form of compatibility & consistency. It's just a different form.

More generally—although the exact wording here is "other raster image format"—we should also consider compatibility & consistency with the web as a whole:
You can use a custom ICC profile in CSS, via the @color-profile at-rule [1].
But you can also use a predefined and enumerated color space, like color(rec2020 ...) [2]. This is very much like H.273.

We don't need every rec2020 image on a page to redefine what rec2020 means.

[1] https://www.w3.org/TR/css-color-4/#at-profile
[2] https://www.w3.org/TR/css-color-4/#predefined

png_cicp_chunk_proposal.md
LarsBrg: ICC is a specific color rendering (one of many)
CICP is a color space, not a color rendering, so it supports a large number of renderings.
Given this, it seems ICC should take priority, as it indicates that a specific rendering has been selected.

png_cicp_chunk_proposal.md
LarsBrg: H.273 is not giving us consistency in rendering of video frames.
CICP is a color space, not a color rendering, so it supports a large number of renderings.
And that's how CICP is used today in video.
- One rendering for scene
- A different rendering for a reference display
- And multiple renderings for displays with non-reference peak nits.

CICP is the best we have today.

png_cicp_chunk_proposal.md
ProgramMax: I mostly follow your suggestion*. But I don't think we can do that. I think the chunk priorities have to match chronological order of chunk standardization.

If someone wanted to specify a color space (which is what the hardware and OSes support), they would have to remove the iCCP chunk. But that leaves a gap between now and when all decoders support cICP. During that time (forever?), the mismatch in decoder support means wider color gamuts become flaky.

Suppose browser A supports cICP and browser B doesn't. Both support iCCP. If the file contains iCCP then the cICP chunk would never be used and would be worthless. If the iCCP chunk is left out, browser A looks correct and browser B does not. So a web dev would likely leave the iCCP chunk in.

It would be difficult to know when it is safe to use only the cICP chunk. Even if browsers A, B, and C support it and that accounts for 99% of a site's users, a web dev would need to actively revisit and weigh trade offs. I fear this would severely hamper cICP support.


* As a browser developer, I'm constrained to what the OS and hardware support/provide. I am able to mark a texture as DCI-P3 in Vulkan but I cannot mark a texture as a specific color rendering, for example. It is easy for me to say "The file says it is in DCI-P3. I'll mark the texture as DCI-P3" and not consider other use cases.

I've been reading a lot to fill my knowledge gaps--ICC specs, wikipedia, etc. My understanding right now is that the color rendering represents the lighting of the scene and the colors that light could produce. IE if my light cannot produce red then a tomato might not show up in my picture. If my understanding is correct, that would make color rendering useful for editing. But for the case of just displaying the image, it isn't useful.

Maybe we could change the wording to something along the lines of "For display purposes, an iCCP chunk can be ignored when a cICP chunk is present." That would allow room for an editor to consider both chunks.

Perhaps on our call tonight you can help me if I misunderstand.

png_cicp_chunk_proposal.md
palemieux: Per the 2021-02-24 meeting:
* there is a desire to support narrow range images, RGB and YUV
* `VideoFullRangeFlag` can signal narrow range for both RGB (the identity matrix is used) and YUV
* all four fields shall be present in a given cICP chunk

png_cicp_chunk_proposal.md
```diff
@@ -13,31 +13,33 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
...
-* VIDFRNG, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **COLPRIMS**, 2 bytes, One of the ColourPrimaries enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **TRANSFC**, 2 bytes, One of the TransferCharacteristics enumerated values specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
+* **VIDFRNG**, 1 byte, Value of the VideoFullRangeFlag specified in Rec. ITU-T H.273 | [ISO/IEC 23091-2]
 
-[ed.: these are inspired from recent JPEG standards that incorporate
-H.273 color space parameters]
+[ed.: these are inspired from recent JPEG standards (eg. JPEG-XL) that incorporate H.273 color space parameters]
 
 [ed.: The MATOEFFS parameter is not included because PNG does not support YCbCr]
```
lrosenthol, line 30: We need to put this back in to support YCbCr

png_cicp_chunk_proposal.md
palemieux: Per 2021-02-24 meeting: remove separation

png_cicp_chunk_proposal.md
palemieux: Per 2021-02-24: we are not concerned with bit-identical serialization of the H.273 parameters

png_cicp_chunk_proposal.md
palemieux: Per 2021-02-24: Use one byte for color primaries, transfer function and matrix coefficient fields, matching H.273.

lrosenthol: Leave a `NOTE` that JPEG standards are 2 bytes vs. 1 here...

png_cicp_chunk_proposal.md
palemieux: Per 2021-02-24: permit cICP and iCCP and defer processor behavior.

png_cicp_chunk_proposal.md
svgeesus: Revising PNG to support YCbCr would be a major undertaking.

palemieux: > Revising PNG to support YCbCr would be a major undertaking.

@svgeesus Can you explain why this is the case if sampling is 4:4:4?

png_cicp_chunk_proposal.md
lrosenthol: I made the change here *however* because of @svgeesus comments about YUV, I am not marking this resolved.  @palemieux

png_cicp_chunk_proposal.md
svgeesus: The [PNG color types](https://www.w3.org/TR/PNG/#6Colour-values) are


PNG image type | Colour type
-- | --
Greyscale | 0
Truecolour | 2
Indexed-colour | 3
Greyscale with alpha | 4
Truecolour with alpha | 6

There is no option in PNG to have YUV. It could be added, but no PNG software would support it and, since it is the image data, it can't be ignored. This would be a major, breaking, revision to a mature and widely deployed specification.

palemieux: @ProgramMax @lrosenthol I suggest that this PR exclude YUV. We can then file a new issue to cover it.

ProgramMax: I have created issue #16 to track/discuss and pull request #17 to implement.

PR #15, opened 2021-03-10
Add HDR compositing of HTMLCanvasElement proposal

ccameron-chromium:
Add a new proposal for HDR compositing of HTMLCanvasElement.

### DIFF
--- a/hdr_html_canvas_element.md
+++ b/hdr_html_canvas_element.md
@@ -1,0 +1,765 @@
+# Canvas High Dynamic Range
+
+## Proposal summary / TLDR
+
+This summary will use terminology from the definitions section below.
+
+### Use cases
+
+There three main classes of HDR use cases that inform this proposal. They are:
+
+* To draw content that can precisely color-match existing SDR content, while allowing to take advantage of HDR headroom.
+  * E.g, adding HDR to an existing SDR application.
+  * E.g, using SDR HTML UI along with an HDR application, with guaranteed color matching.
+* To display PQ encoded HDR images that are drawn to a canvas
+  * E.g, displaying PQ content.
+  * E.g, working in physical luminance.
+* To display HLG encoded HDR images that are drawn to a canvas
+  * E.g, displaying HLG content.
+  * E.g, using a fixed signal range that maps to the full display device luminance range.
+
+### Constraints
+
+There exist the following constraints.
+
+* The exact maximum luminance of the output display is not known and not knowable.
+  * The [CSS Media Queries Level 5 Specification](https://www.w3.org/TR/mediaqueries-5/#valdef-media-dynamic-range-high) allows the application to query the ``'dynamic-range'``. The resulting values are ``'standard'`` and ``'high'``.
+  * The value changes over time.
+  * The exact value is a fingerprinting vector.
+* The exact number of nits of SDR content is also not known and not knowable.
+  * On macOS, it is always 100.
+  * On Windows, it depends on a user slider setting.
+  * The exact value is a fingerprinting vector (again).
+
+Because these values are not known, it is not possible for the application provide quantities related to display light.
+
+### Proposed solution
+
+The solution that we propose is to:
+
+* Introduce the term device independent light.
+* Introduce new color spaces and precisions that are useful for HDR.
+  * These color spaces are best interpreted as signals, which have a well defined conversion into device independent light.
+* Introduce an HTMLCanvasElement method through which an HDR compositing mode may be specified.
+  * The compositing mode corresponds to a class of opto-optical transfer functions that are applied to the device independent light defined by the canvas' buffer's contents to determine the canvas' final display light.
+  * We introduce three HDR compositing modes, matching the three main use cases.
+  * Note that these are modes of compositing a canvas, are independent of the canvas' color space, and may be changed between dynamically.
+* Define mappings from SDR, PQ, and HLG signals into display independent light.
+  * Mappings are chosen to complement the HDR compositing modes.
+  * Remaining free parameters in the HLG mapping are chosen to ensure a smooth fallback when HDR is disabled.
+  * Remaining free parameters in the PQ mapping are chosen to make math easier.
+* If the application can overcome the fingerprinting limitations (e.g, by just asking the user), any desired behavior can be accomplished, using appropriate math.
+
+## Definitions
+
+In this section we provide definitions for the terminology that is used throughout the document.
+These definitions focus on being precise about brightness, and will not be as precise about color.
+
+### Display light, scene light, device independent light, and signal
+
+#### Display light
+
+Display light is the light that is emitted by a display device (e.g, a television) at each pixel.
+Display light has the property that it is limited by a maximum and minimum value that the display device can produce.
+
+Display light is a precise physical quantity, namely, luminance in nits (nits are cd/m^2).
+
+Display light is sometimes treated as being in the interval [0, 1], with 1 representing the maximum luminance the device can produce, and 0 possibly representing the minimum luminance that the device can produce.
+In these cases display light is sometimes referred to as being "linear", because nothing more than a linear (or more precisely, affine) transformation has been applied to it.
+We will not use this convention.
+
+#### Relative scene linear light
+
+Relative scene linear light is linearly proportional to the physically measurable light in a scene being captured.
+The precise scaling factor used in mapping from physically measurable light to relative scene linear light depends on the particular capture setup, including such variables as camera exposure.
+The scaling of relative scene linear light is such that values of scene linear light are limited to the interval [0, 1].
+
+This document uses the term "scene light" as shorthand for "relative scene linear light".
+TODO: Fix this if it is a source of confusion.
+
+#### Device independent light
+
+Device independent light is a space into which all content (SDR, PQ, and HLG) has a well-defined and simple mapping.
+
+The reason for the existence of device independent light is that there needs to exist a space that does not depend on the display device, in which content is clearly defined.
+
+Device independent light is extremely similar to relative scene linear light.
+Earier versions of this document used the term scene light (meaning relative scene linear light) instead of device independent light, but this was too confusing because of the issues of artistic adjustments and the undefined nature of scene light for SDR content.
+
+#### Signal
+
+A signal is a digial encoding of light.
+
+An example of a signal is the pixel values in an sRGB image.
+Other examples include the pixel values of PQ or HLG images.
+
+Most signals take values in the [0, 1] interval and are represented using fixed-point (e.g, the usual 8-bit encoding of an sRGB image).
+In this document, we will be allowing for signals that take any real value, and are represented using floating-point.
+
+### Transfer functions
+
+#### Opto-electronic transfer function (OETF)
+
+An opto-electronic transfer function (OETF) is a function that transforms scene light (or sometimes modified scene light) to a signal.
+
+An example of an OETF is the HLG OETF, specified in the second row of Table 5 of the [BT.2100 specification](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-2-201807-I!!PDF-E.pdf).
+
+Another example of an OETF is the PQ OETF, specified in the final row of Table 4 of the [BT.2100 specification](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-2-201807-I!!PDF-E.pdf).
+
+A non-example of an OETF is the [sRGB function](https://www.w3.org/Graphics/Color/srgb) ``f(x)=(x<0.0031308) ? 12.92*x : 1.055*pow(x,1/2.4)-0.055``.
+The domain of this OETF-like thing is the interval [0, 1], with diffuse white at 1.
+The range (signal) is standard sRGB pixel values in the interval [0, 1].
+This is not an OETF because sRGB only specifies display light for sRGB signals, it does not define a relationship to scene light (in other words, the domain of this function is not scene light).
+
+#### Opto-optical transfer function (OOTF)
+
+An opto-optical transfer function (OOTF) is a function that transforms scene light to display light.
+
+It is the responsibility of the OOTF to reproduce the appearance of the scene, within the capabilities of the display device, and subject to viewer's environment.
+
+The HLG OOTF is parameterized by the maximum luminance of the display device. It maps a fixed range of scene light to cover the full range (up to maximum luminance) of the display device.
+
+The reference PQ OOTF is close to the identity function, with scene light almost equaling display light.
+In practice (that is, when applied to a non-reference display), the PQ OOTF is parameterized by content metadata and the capabilities of the display device.
+
+#### Electro-optical transfer function (EOTF)
+
+An electro-optical transfer function (EOTF) is a function that transforms signal to display light, and can be computed as the OOTF applied to the inverse-OETF applied to a signal.
+We are including this definition for completeness, but this document will be written in terms of OETFs and OOTFs.
+
+#### Device independent electro-optical transfer function (DI-EOTF)
+
+A device independent electro-optical transfer function (DI-EOTF) is a function that transforms a signal to display independent light.
+
+In this proposal we will define the DI-EOTF for sRGB, PQ, and HLG signals.
+
+#### Device independent opto-optical transfer function (DI-OOTF)
+
+A device independent opto-optical transfer function (DI-OOTF) is a function that transforms display independent light to display light.
+
+In this proposal we will define three classes of DI-OOTFs that the application may use to display HDR content.
+
+### Application versus browser
+
+In this document, the term application will refer to the web application that does not have direct access to the operating system.
+This is in contrast with the browser, which is a native application with direct access to the operating system.
+
+## New color spaces and storage formats
+
+The existing CanvasColorSpaceProposal has been narrowed down to supporting just ``'srgb'`` and ``'display-p3'``.
+It no longer covers adding other color spaces, or changing buffer formats.
+
+### Color spaces
+
+The pixels values in a canvas back buffer are in a specific color space.
+The default color space is sRGB.
+
+This new proposal expands the set of supported color spaces to include:
+
+* ``'srgb'``
+* ``'display-p3'``
+* ``'srgb-linear'``
+* ``'display-p3-linear'``
+* ``'rec2020-linear'``
+
+#### Interpretation in terms of signal
+
+The canvas back buffer is best understood as a signal, or an encoded version of device independent light.
+
+The color space determines this encoding.
+In particular, the color space specifies a DI-EOTF that maps the signal in from device independent light.
+For all of the above-listed color spaces, this DI-EOTF and its inverse are defined for all real numbers.
+
+The pixel value of (1,1,1) in all color spaces corresponds diffuse white.
+
+(The reader may note that color primaries are not being attended to in this section. Their neglect will continue.)
+
+#### Relationship to BT.2100 floating-point signal representation.
+
+When a canvas has color space ``'rec2020-linear'``, the back buffer's pixel values are precisely the scene-referred signal referred to in Table 10 of the [BT.2100 specification](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-2-201807-I!!PDF-E.pdf).
+
+### Storage formats
+
+This proposal includes adding 16-bit floating-point as a supported storage format for 2D Canvas.
+
+ImageData will also be updated to support 16-bit fixed-point and 32-bit floating-point.
+
+## HDR compositing modes
+
+In this section we go over the three HDR compositing modes that we propose to make available.
+These are compositing modes, which means that their effects are visible to the user via the display device, but are not visible to the application itself.
+
+As discussed in the previous section, the contents of a canvas' buffer may be interpreted as a signal that has a well-defined mapping into display independent light.
+The compositing mode selects which class of device independent opto-optical transfer function is to be applied to that display independent light to convert it to display light produced by the display device.
+It is the responsibility of the browser, the operating system, the graphics hardware, and the display hardware to effect the application-selected OOTF.
+
+In this section, we will define how we will map PQ and HLG signals to our set of defined color spaces.
+These are necessarily raw signal-to-signal mappings, done without knowledge of the display device or viewing environment.
+
+### Mode 1: Extended-SDR mode or SDR-relative luminance (for matching SDR colors)
+
+The requirement of this mode is that all SDR colors inside the HDR Canvas match exactly their appearance in an SDR canvas.
+
+In this mode, color values outside of the range of [0, 1] may be used to represent luminance beyond the SDR range.
+The exact luminance of such a color value is expressed relative to the maximum SDR luminance, rather than in absolute nits.
+For example, the color ``color(srgb-linear 2 2 2)`` is exactly twice as bright as ``color(srgb 1 1 1)``, but it not known how many nits that is.
+
+#### Metadata and tone mapping
+
+There is no limit on the maximum luminance that can be expressed by a pixel value in this mode.
+If no additional metadata is provided, then all pixels will be clamped to the display device's maximum luminance.
+
+To avoid aggressive clamping, metadata may be provided.
+This metadata consists of the scene's maximum luminance, as a multiple of the SDR luminance.
+A combination of the browser, operating system, and display device are responsible for mapping the canvas' content into the display device's capabilities.
+
+Note that this mapping has the very severe restriction that it must not alter SDR colors found inside the HDR canvas (in other words, the mapping must be the identity on all SDR values).
+
+Also note that no mastering primaries or white point are specified, because there is no way to incorporate that data into a mapping that remains the identity on all SDR values.
+
+#### Remarks on operating system interaction
+
+On Windows, this mode opts the canvas in to being affected the SDR slider (like any non-HDR canvas would be).
+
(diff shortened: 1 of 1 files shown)


simontWork (commented):
Added some comments on clamping, low power devices, definitions, OOTF.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+The second reason to prefer normalized scene light is that the non-normalized representation will not behave well if applied to a non-SDR canvas. In particular, it will saturate and start to clamp at a signal value of 0.5.
+A defining property of HLG is the smooth transition between SDR and HDR, this clamping is not a smooth transition, and so something different would need to be done in non-SDR canvases.
+That would end up with a different scene light representation of an HLG image depending on whether it is destined for an HDR or SDR canvas, which is a mess from the perspective of the application and the browser.
+
+### Mode 0: SDR
+
+For completeness, there does exist one more HDR mode, the no-HDR HDR mode (which is the current default behavior).
+
+In this mode, scene light is clamped to the range [0, 1], and the SDR OOTF is applied.
+This is done by having the native application (the browser in this case) provides the operating system with a buffer in a specified color space.
```
chrisn, line 321: Typo: provides -> provide

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+In this example, the application does not want color matching with the HTML UI.
+The application views the HTML UI as being regular SDR content that is sitting on top of an HDR canvas.
+
+* The UI should be displayed however the underlying OS displays SDR content.
+* The HDR canvas should be displayed however the underlying OS displays HDR content.
+
+The application in this case will want ``'pq-compatible'``.
+
+```javascript
+    canvas.configureHDR('pq-compatible'});
```
chrisn, line 481: Typo: extra '}' (and similar in the other examples)

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+  </body>
+  </html>
+```
+
+#### Example 4B: Displaying an HLG image in a 2D Canvas NOT as intended
+
+In this example, we make a mistake, and use ``'extended'`` in the above exmple.
+
+```javascript
+  canvas.configureHDR('extended'});
```
chrisn, line 715: Typo: extra '}'

chrisn (changes requested):

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+In this example, the application wants color matching with the HTML UI.
+The most concrete example of this would be a situation where the application uses the HTML color picker element to select colors to use inside the canvas.
+
+In this case, the application will want to use the default mode of ``'extended'``.
+
+```javascript
+    canvas.configureHDR('extended');
+
+    // This green will match the color in MyGreenUI.
+    var context = canvas.getContext('2d', precision='float16');
```
chrisn, line 451: ```var context = canvas.getContext('2d', { precision: 'float16' });```

(and similar in the other examples)

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+To enable the configuration of HDR compositing.
+
+```idl
+  partial interface HTMLCanvasElement {
+    void configureHDR(HTMLCanvasCompositingMode mode, optional HTMLCanvasHDRMetadata metadata);
+  }
+
+  enum HTMLCanvasHDRCompositingMode {
+    'disabled',
+    'extended', // Call this "relative-luminance"?
```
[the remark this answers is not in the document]
svgeesus: I think *extended* means in general allowing component values greater than 100% and less than 0%, to cope with out of gamut colors. So therefore it is not specific to sRGB.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+  * The compositing mode corresponds to a class of opto-optical transfer functions that are applied to the scene light defined by the canvas' buffer's contents to determine the canvas' final display light.
+  * We introduce three HDR compositing modes, matching the three main use cases.
+  * Note that these are modes of compositing a canvas, are independent of the canvas' color space, and may be changed between dynamically.
+* Define mappings from PQ and HLG signals into scene light, and, by extension, the various provided color spaces.
+  * Mappings are chosen to complement the HDR compositing modes.
+  * Remaining free parameters in the HLG mapping are chosen to ensure a smooth fallback when HDR is disabled.
+  * Remaining free parameters in the PQ mapping are chosen to make math easier.
+* If the application can overcome the fingerprinting limitations (e.g, by just asking the user), any desired behavior can be accomplished, using appropriate math.
+
+## Definitions
```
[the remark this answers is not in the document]
ccameron-chromium: In that case we should link to the definitions that the ITU provides -- please provide links where the ITU defines these terms and I'll include them.

I haven't found any place where ITU defines what "scene light" is supposed to mean (is it in nits, is it normalized, is it only the measurable light or can it have alterations done, etc, and the usage of the term in BT2100 is flexible).

hdr_html_canvas_element.md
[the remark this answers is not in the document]
ccameron-chromium: Correct, the "extended" refers to "extending the SDR range", and can be applied to any color space (srgb, linear-srgb, display-p3, linear-rec2020, etc).

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,633 @@
+# Canvas High Dynamic Range
+
+## Proposal Summary / TLDR
+
+### Use Cases
+
+There three main classes of HDR use cases that inform this proposal. The are:
```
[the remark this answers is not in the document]
ccameron-chromium: Thanks for these corrections, I think I've made all of them. (With every revision I cut and paste things around and re-mangle the sentences, and my brain seems to auto-correct things during proofreading)

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+
+### Mode 3: HLG transformation of scene light
+
+The goal of this mode is to allow faithful display of HLG encoded content.
+An HLG image drawn to a canvas that is composited in this mode will appear the same as that image when drawn by an ``img`` element on the page.
+
+That is not the only use of this mode.
+An application that wishes to produce pixel values in the interval [0, 1] in scene light space and have that interval gracefully map to the full display's luminance range (be that range HDR or SDR) would find this mode a natural fit.
+
+#### Background on converting HLQ signals to display light
```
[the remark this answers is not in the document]
ccameron-chromium: Thanks -- fixed.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+#### Example 1B: Same as 1A, but using WebGL
+
+This is the same as the above example, but the application is using WebGL.
+
+```javascript
+    canvas.configureHDR('extended');
+
+    // This green will match the color in MyGreenUI.
+    var gl = canvas.getContext('webgl2');
+    gl.configureSwapChain({ format: gl.RGBA16F, colorSpace: 'srgb-linear' });
```
kenrussell, line 465: The exact API here is still in discussion. It's likely to be:
```
gl.drawingBufferStorage(gl.RGBA16F, canvas.width, canvas.height);
gl.colorSpace = 'srgb-linear';
```

See https://github.com/KhronosGroup/WebGL/pull/3222 .

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+    // Enable HDR for the canvas. Use the default mode of 'extended',
+    // because that will match the existing application behavior.
+    canvas.configureHDR('extended');
+
+    var gl = canvas.getContext('webgl2');
+
+    // Configure the swap chain to be floating-point. Leave the color space as
+    // the default of 'srgb'. This will leave all colors in [0, 1] unchanged,
+    // but allow for specifying colors outside of the range of [0, 1].
+    gl.configureSwapChain({ format: gl.RGBA16F });
```
kenrussell, line 568: Needs updating per above comment about exact form of the WebGL API.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+  <html>
+  <body>
+  <canvas id='MyCanvas' width=500px height=500px
+          style='position:absolute; width: 100%; height:100%;'></canvas>
+  <script>
+    var canvas = document.getElementById('MyCanvas');
+    canvas.configureHDR('extended');
+    var context = canvas.getContext('2d', { precision: 'float16' });
+
+    function animateBrigtenedUI(timeInSections) {
```
kenrussell, line 517: Small typo: Brightened

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+
+For PQ images, the behavior that falls out is for the images to be clamped beyond 100 nits.
+
+## Proposed API outline
+
+To enable the configuration of HDR compositing.
+
+```idl
+  partial interface HTMLCanvasElement {
+    void configureHDR(HTMLCanvasCompositingMode mode, optional HTMLCanvasHDRMetadata metadata);
```
kenrussell, line 335: Should this be specified for OffscreenCanvas as well? Presumably that would affect both ImageBitmap objects created via `transferToImageBitmap`, and if the OffscreenCanvas was controlling an HTMLCanvasElement, then would also affect the canvas, similarly to how resizing an OffscreenCanvas eventually causes the HTMLCanvasElement to resize.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+  }
+
+  enum HTMLCanvasHDRCompositingMode {
```
kenrussell, line 338: I think this needs `[Exposed=(Window,Worker)]`.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+    'disabled',
+    'extended', // Call this "relative-luminance"?
+    'pq-compatible', // Call this "absolute-luminance" or "physical-luminance"?
+    'hybrid-log-gamma',
+  }
+
+  // TODO: This feels sloppy to have all HDR parameters be in a single
+  // dictionary.
+  dictionary HTMLCanvasHDRMetadata {
```
kenrussell, line 347: Think this similarly needs the `Exposed` attribute.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+    float? bluePrimaryY;
+    float? whitePointX;
+    float? whitePointY;
+  }
+```
+
+To enable 2D Canvas to use more than 8 bits per pixel.
+
+```idl
+  enum CanvasStorageFormat {
```
kenrussell, line 373: Here and below, believe these all need the `Exposed` extended attribute.

kenrussell (approved):
Reviewed this mainly from the WebGL standpoint; it looks good to me overall. A few relatively minor comments throughout.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+
+An example of an OETF is the [sRGB function](https://www.w3.org/Graphics/Color/srgb) ``f(x)=(x<0.0031308) ? 12.92*x : 1.055*pow(x,1/2.4)-0.055``.
+The domain of this OETF (scene light) is the interval [0, 1], with diffuse white at 1 (which is what we adopted as convention).
+The range (signal) is standard sRGB pixel values in the interval [0, 1].
+
+Another example of an OETF is the HLG OETF, specified in Table 5 of the [BT.2100 specification](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-2-201807-I!!PDF-E.pdf).
+For this OETF, the domain (scene light) is normalized to the interval [0, 1], with maximum brightness at 1 (which we will treat as matching our convention).
+
+Another example of an OETF is the PQ OETF, specified in Table 4 of the [BT.2100 specification](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-2-201807-I!!PDF-E.pdf).
+For this OETF, the domain (scene light) is luminance values in nits (which does not not match our convention, and we will discuss how to match this in detail).
```
simontWork, line 105: Is this correct? In table 4, the OETF maps relative scene linear light into the non-linear PQ signal value.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+
+#### Opto-optical transfer function (OOTF)
+
+An opto-optical transfer function (OOTF) is a function that transforms scene light to display light.
+
+It is the responsibility of the OOTF to reproduce the appearance of the scene, within the capabilities of the display device, and subject to viewer's environment.
+
+The HLG OOTF is parameterized by the maximum luminance of the display device. It maps a fixed range of scene light to cover the full range (up to maximum luminance) of the display device.
+
+The PQ OOTF can be thought of as the identity function, with scene light equaling display light.
```
simontWork, line 115: Is this correct? BT.2100 defines the OOTF as approx. a gamma 1.2 applied on RGB.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+
+Transforming from HLG signal to raw luminance has three steps.
+
+The first step is to convert from the HLG signal into scene light. This is done by the inverse-OETF function.
+The function's domain is [0, 1], and its range is normalized to [0, 1], with 0.5 mapping to 1/12.
+(Sometimes this range is normalized to [0, 12], but we do not use that convention).
+This step is independent of the output display.
+
+The second step is to apply the normalized OOTF to transform scene light into normalized display light.
+This function is a gamma ramp, parameterized by the maximum luminance of the display, with a domain and range that are [0, 1].
```
simontWork, line 284: Gamma ramp applied on Luminance, not RGB.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+The nits specified by PQ image will match the nits displayed on the screen, as much as is allowed by the underlying operating system and the physical display.
+
+```xml
+  <html>
+  <body>
+  <canvas id='MyCanvas' width=2048px height=858px
+          style='position:absolute; width: 95%;'></canvas>
+  <script>
+    var canvas = document.getElementById('MyCanvas');
+    canvas.configureHDR('pq-compatible'});
```
chrisn, line 606: Typo: extra `}`

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+  </body>
+  </html>
+```
+
+#### Example 3B: Displaying a PQ image in a 2D Canvas NOT as intended
+
+In this example, we make a mistake, and use ``'extended'`` in the above exmple.
+
+```javascript
+  canvas.configureHDR('extended'});
```
chrisn, line 625: Typo: extra `}`

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+We set the canvas to use the appropriate compositing mode.
+
+```xml
+  <html>
+  <body>
+  <canvas id='MyCanvas' width=2048px height=858px
+          style='position:absolute; width: 95%;'></canvas>
+  <script>
+    var canvas = document.getElementById('MyCanvas');
+    canvas.configureHDR('hybrid-log-gamma'});
```
chrisn, line 696: Typo: extra `}`

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+```javascript
+  image.onload = function() {
+    context.drawImage(image, 0, 0, 2048, 858);
+    context.font = "128px Arial";
+    context.fillStyle = 'white';
+    context.fillText('Hello, I am a subtitle!', 400, 800);
+  }
+```
+
+It is a convention in HLG to use a signal value of 0.75 as diffuse wwhite.
```
chrisn, line 738: Typo: white

chrisn (changes requested):

simontWork (commented):
Added a couple more comments.

hdr_html_canvas_element.md
ccameron-chromium: Going to defer fixing Exposed for the moment (my understanding of it isn't deep enough)

hdr_html_canvas_element.md
ccameron-chromium: Fixed

hdr_html_canvas_element.md
ccameron-chromium: Fixed

hdr_html_canvas_element.md
ccameron-chromium: Fixed

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
+# Canvas High Dynamic Range
+
+## Proposal summary / TLDR
+
+This summary will use terminology from the below definitions section.
```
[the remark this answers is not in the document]
ccameron-chromium: Fixed

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,633 @@
...
+This is an abbreviated summary of [PQ to HLG Transcoding](https://www.bbc.co.uk/rd/sites/50335ff370b5c262af000004/assets/592eea8006d63e5e5200f90d/BBC_HDRTV_PQ_HLG_Transcode_v2.pdf).
+
+Transforming from HLG signal to raw luminance has three steps.
+
+The first step is to convert from the HLG signal into scene light. This is done by the inverse-OETF function.
+The function's domain is [0, 1], and its range is normalized to [0, 1] (with 0.5 mapping to 1/12).
+This step is independent of the output display.
+
+The second step is to apply the OOTF to transform scene light into display light.
+This function is a gamma ramp, parameterized by the maximum luminance of the display, with a domain and range are [0, 1].
```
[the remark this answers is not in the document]
ccameron-chromium: Changed slightly

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,633 @@
...
+
+* For input PQ images drawn to canvases, we have to map nits to color values.
+* For outputting canvases in this compositing mode, we have to map color values to nits.
+
+We choose to assign a the same mapping for both of these, so that simply drawing a PQ image to a canvas that is composited in this mode will behave as expected.
+The alternative would be to allow the user to parameterize the output mode (as compositing metadata) and the input mode (via ImageBitmap).
+Such parameterization allows for the user to create an unexpected mismatch, which would be undesirable.
+
+The mapping we decide is that the color ``color(srgb 1 1 1)`` is to map to 100 nits.
+The motivation is that it's reasonable and that it the makes math easier for anyone who desires a different value (and many applications will want many other values).
```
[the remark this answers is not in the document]
ccameron-chromium: Fixed

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,633 @@
...
+
+#### Assigning pixel values to nits specified by PQ signals
+
+In this mode, we assign a number of nits to colors.
+We have to do this in two places.
+
+* For input PQ images drawn to canvases, we have to map nits to color values.
+* For outputting canvases in this compositing mode, we have to map color values to nits.
+
+We choose to assign a the same mapping for both of these, so that simply drawing a PQ image to a canvas that is composited in this mode will behave as expected.
```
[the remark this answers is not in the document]
ccameron-chromium: Reworded most of the section.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,633 @@
...
+
+There three main classes of HDR use cases that inform this proposal. The are:
+
+* To draw content that can precisely color-match existing SDR content, while allowing to take advantage of HDR headroom.
+  * E.g, adding HDR to an existing SDR application.
+  * E.g, using SDR HTML UI along with an HDR application, with guaranteed color matching.
+* To draw PQ encoded HDR images as they are intended, displaying the specified nits.
+  * E.g, displaying PQ content.
+  * E.g, working in physical luminance.
+* To draw HLG encoded HDR images as the are intended, mapping the HLG signal space to the full display device luminance range.
```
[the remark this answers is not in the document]
ccameron-chromium: Reworded most of the section.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+A more complete treatment of this topic may be found in the very readable [PQ to HLG Transcoding](https://www.bbc.co.uk/rd/sites/50335ff370b5c262af000004/assets/592eea8006d63e5e5200f90d/BBC_HDRTV_PQ_HLG_Transcode_v2.pdf) or in the [BT.2100 specification](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-2-201807-I!!PDF-E.pdf).
+
+Transforming from HLG signal to raw luminance has three steps.
+
+The first step is to convert from the HLG signal into scene light. This is done by the inverse-OETF function.
+The function's domain is [0, 1], and its range is normalized to [0, 1], with 0.5 mapping to 1/12.
+(Sometimes this range is normalized to [0, 12], but we do not use that convention).
+This step is independent of the output display.
+
+The second step is to apply the normalized OOTF to transform scene light into normalized display light.
```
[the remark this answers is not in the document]
ccameron-chromium: On OSes that accept an HLG signal (or linearized HLG signal), we will want to do exactly that -- the browser will apply no OOTF. The buck is then passed to the OS, which may pass it to the monitor or TV.

But, not all OSes accept HLG signals, so then the buck stops with the browser. In that case, if the browser is to allow HLG content, then the browser must be the one do whatever transcoding is needed to get the intended HLG content into whatever format the OS accepts.

hdr_html_canvas_element.md
```diff
@@ -0,0 +1,744 @@
...
+* To display HLG encoded HDR images that are drawn to a canvas
+  * E.g, displaying HLG content.
+  * E.g, using a fixed signal range that maps to the full display device luminance range.
+
+### Constraints
+
+There exist the following constraints.
+
+* The exact maximum luminance of the output display is not known and not knowable.
+  * The [CSS Media Queries Level 5 Specification](https://www.w3.org/TR/mediaqueries-5/#valdef-media-dynamic-range-high) allows the application to query the ``'dynamic-range'``. The resulting values are ``'standard'`` and ``'high'``.
```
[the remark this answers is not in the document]
ccameron-chromium: This and pal's comments are both suggesting a need for a PQ or HLG color space (which would probably pair with 10-bit-per-channel formats).

hdr_html_canvas_element.md
ccameron-chromium: Suppose we have a canvas that has had an image drawn into it (SDR, PQ, or HLG) and nothing else. The goal that I have in mind is for the result of transferToImageBitmap to have the same contents as an ImageBitmap created from the original image would. Because of how everything is defined in "device independent light" (in the new upload), this ends up being the case -- no compositing mode information needed(!).

The only place where this could fall apart is if the ImageBitmap can be assigned to an  tag. In that case, the resulting  would not look the same as the canvas that it came from. Is that an option?

hdr_html_canvas_element.md
ccameron-chromium: This is a good example of the definition problem that I was discussing. (Short answer: "you're right").

What is "relative scene linear light"? BT2100 uses it, but gives no definition. From its use, and with some detective work, we can piece together what might be a definition that is consistent with its usage.

In Table 4 that "Input signal to PQ opto-optical transfer function (OOTF)" is "Scene linear light. The OOTF maps relative scene linear light to display linear light." Well, which is it, "scene linear light" or "relative scene linear light"? They're different words, so presumably they have different meanings? What does "relative" mean here? Why is the word "linear" in there?

The next row in Table 4 contains the symbols "OOTF[E]", so I guess this means that E is the input signal to the OOTF which we established is either "scene linear light" or "relative scene linear light". (let's just go with "relative scene linear light"). And then E is defined as "E is the signal determined by scene light and scaled by camera exposure". No definition for "scene light" either, but at least that one is intuitive.

Table 5 has the same text, except defines E as "E is a signal for each colour component ... proportional to scene linear light normalized to the range [0:1]".

If we go back to Table 4, we can see that in the "Reference PQ OOTF" row, "E′ = G_{709}[E]" is defined only for values of E in the half-open interval [0, 1). So that supports the hypothesis that relative scene linear light is limited to the unit interval (perhaps closed or perhaps half-open).

Putting all of this together, we could say that "Relative scene linear light (denoted by E in this document), is a value in the interval [0, 1] that is linearly proportional to the physically measurable light in the scene being captured. The precise scaling factor used in mapping from physically measurable light to relative scene linear light depends on the particular capture setup, including such variables as camera exposure".

This is the sort of definition that I want to put at the top of the proposal -- I want it to be readable to people who have a basic understanding of compute graphics and the physics of light, but who are not steeped in broadcast terminology.

ccameron-chromium:
Merging this with a clarification about "display independent light"

hdr_html_canvas_element.md
kenrussell: > Suppose we have a canvas that has had an image drawn into it (SDR, PQ, or HLG) and nothing else. The goal that I have in mind is for the result of transferToImageBitmap to have the same contents as an ImageBitmap created from the original image would. Because of how everything is defined in "device independent light" (in the new upload), this ends up being the case -- no compositing mode information needed(!).
> 
> The only place where this could fall apart is if the ImageBitmap can be assigned to an  tag. In that case, the resulting  would not look the same as the canvas that it came from. Is that an option?

(Sorry, which tag? &lt;img&gt;? I think Github's editor deleted the angle-bracketed-tag name you referenced.)

HTMLImageElement's src attribute can not be set to an ImageBitmap. ImageBitmap in general can only be used with the imperative JavaScript APIs like the canvas rendering contexts, OffscreenCanvas, etc.

Issue #18, opened 2021-03-16
Specify expected behavior in presence of both cICP and ICC profile information


palemieux:
> A PNG MAY contain both a cICP chunk and an iCCP chunk. If a PNG decoder detects the presence of both a cICP and a iCCP chunk, the behavior is undefined.

The behavior should be defined.

ProgramMax:
I've been thinking about this.

For HDR content, the existing iCCP chunk is insufficient. In that case, we wouldn't want to prioritize iCCP over cICP.
(Is the existing iCCP chunk sufficient for WC?)

However, if we were using ICC v4 profiles (instead of iCCP's v2), maybe it would have been sufficient.
So maybe we want iCC4 (new chunk, can rename) > cICP > iCCP.

What I just described is specific to HDR though. For WC, I can understand that iCCP may be sufficient and is more specific. I think this is where the confusion and divide came from.

I don't think we want a rule like "If the cICP chunk describes a HDR color space, it takes priority. If not, the iCCP chunk takes priority." Even if we carefully defined what counts as HDR, this feels like the wrong approach.

Instead, if the existing iCCP chunk is sufficient for WC (or any other non-HDR color), I think the user/creation program could simply provide only the iCCP chunk and not the cICP chunk. That way it effectively takes priority. They have no need for the cICP chunk so why waste bytes?

I believe the user would only provide the cICP chunk when the iCCP chunk is insufficient. In which case, it should take priority. And later when we have support for ICC profiles that are sufficient for HDR, it can take priority over cICP.

@LarsBrg Do you agree with this? Am I missing anything?

If this looks good, we'll want to think about the iCC4 chunk enough to make sure we aren't making a mistake there.

LarsBrg:
Looks good. See some comments inline.

 

From: Chris Blume ***@***.***>

For HDR content, the existing iCCP chunk is insufficient. In that case, we wouldn't want to prioritize iCCP over cICP.
(Is the existing iCCP chunk sufficient for WC?)

WC == Wide color gamut? Yes, ICC V2 handles wide color gamut such as Rec.2020.

Yes, AFAIK HDR requires ICC V4. I make such profiles, and I haven’t seen any ICC V2 profiles made for HDR.

However, if we were using ICC v4 profiles (instead of iCCP's v2), maybe it would have been sufficient.
So maybe we want iCC4 (new chunk, can rename) > cICP > iCCP.

[...]

PR #19, opened 2021-03-17
Proposal for new iCCN tag for PNG

lrosenthol:
Proposal to add a new version of the ICCP tag to address #18

### DIFF
--- a/iCCN_for_PNG.md
+++ b/iCCN_for_PNG.md
@@ -1,0 +1,26 @@
+# iCCN
+
+## iCCN Embedded ICC profile (updated)
+The four-byte chunk type field contains the decimal values
+
+`105 67 67 80`
+
+The iCCN chunk contains:
+
+| Profile name |  1-79 bytes (character string)
+| Null separator |	1 byte (null character)
+| Compression method |	1 byte
+| Compressed profile |	n bytes
+
+The profile name may be any convenient name for referring to the profile. It is case-sensitive. Profile names shall be encoded as UTF-8. Leading, trailing, and consecutive spaces are not permitted. The only compression method defined in this International Standard is method 0 (zlib datastream with deflate compression, see 10.3: Other uses of compression). The compression method entry is followed by a compressed datastream of an ICC profile as defined in [ICC] or [ICC-2001]. The ICC profile shall either be an output profile (Device Class = `prtr`) or a monitor profile (Device Class = `mntr`). Decompression of this datastream yields the embedded ICC profile.
+
+If the iCCN chunk is present, the image samples conform to the colour space represented by the embedded ICC profile. The colour space of the ICC profile shall be an RGB (`RGB `) colour space for colour images (PNG colour types 2, 3, and 6), or a greyscale (`GRAY`) colour space for greyscale images (PNG colour types 0 and 4). A PNG encoder that writes the iCCN chunk is encouraged to also write `gAMA` and `cHRM` chunks that approximate the ICC profile, to provide compatibility with applications that do not use the iCCN chunk. When the iCCN chunk is present, PNG decoders that recognize it and are capable of colour management shall ignore any `gAMA`, `cHRM`, and `CiCP` chunks and use the iCCN chunk instead and interpret it according to [ICC] or [ICC-2001] as appropriate. PNG decoders that are used in an environment that is incapable of full-fledged colour management should use the gAMA and cHRM chunks if present.
+
+A PNG datastream shall contain at most one embedded profile, whether specified explicitly with an `iCCP` or `iCCN` chunk or implicitly with an `sRGB` chunk.
+
+## Normative References
+
+[ICC] ISO 15076-1:2010, Image technology colour management – Architecture, profile format and data structure — Part 1: Based on ICC.1:2010
+
+[ICC-2001] Specification ICC.1:2001-04, File Format for Color Profiles
+
--- a/iCCN_for_PNG.md
+++ b/iCCN_for_PNG.md
@@ -5,16 +5,23 @@ The four-byte chunk type field contains the decimal values
 
 `105 67 67 80`
 
-The iCCN chunk contains:
+The `iCCN` chunk contains:
 
 | Profile name |  1-79 bytes (character string)
 | Null separator |	1 byte (null character)
 | Compression method |	1 byte
 | Compressed profile |	n bytes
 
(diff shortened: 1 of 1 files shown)


palemieux (changes requested):
Can this made as a PR against https://github.com/w3c/ColorWeb-CG/blob/master/png_cicp_chunk_proposal.md ?

iCCN_for_PNG.md
```diff
@@ -0,0 +1,26 @@
...
+
+A PNG datastream shall contain at most one embedded profile, whether specified explicitly with an `iCCP` or `iCCN` chunk or implicitly with an `sRGB` chunk.
+
+## Normative References
+
+[ICC] ISO 15076-1:2010, Image technology colour management – Architecture, profile format and data structure — Part 1: Based on ICC.1:2010
+
+[ICC-2001] Specification ICC.1:2001-04, File Format for Color Profiles
```
svgeesus, line 25: This should be

[ICC-2010] Specification ICC.1:2010-12 (Profile version 4.3.0.0) Image technology colour management - Architecture, profile format, and data structure 

also, unless the intent is specifically to exclude iccMAX, the following should be added:

[ICC-2019] Specification ICC.2:2019 (iccMAX) Image technology colour management - Extensions to architecture, profile format, and data structure 

Related, [PNG erratum](https://github.com/w3c/PNG-spec/issues/22) to add these to the PNG spec which, due to the age of the specification, only references ICC v.2

iCCN_for_PNG.md
```diff
@@ -0,0 +1,26 @@
...
+`105 67 67 80`
+
+The iCCN chunk contains:
+
+| Profile name |  1-79 bytes (character string)
+| Null separator |	1 byte (null character)
+| Compression method |	1 byte
+| Compressed profile |	n bytes
+
+The profile name may be any convenient name for referring to the profile. It is case-sensitive. Profile names shall be encoded as UTF-8. Leading, trailing, and consecutive spaces are not permitted. The only compression method defined in this International Standard is method 0 (zlib datastream with deflate compression, see 10.3: Other uses of compression). The compression method entry is followed by a compressed datastream of an ICC profile as defined in [ICC] or [ICC-2001]. The ICC profile shall either be an output profile (Device Class = `prtr`) or a monitor profile (Device Class = `mntr`). Decompression of this datastream yields the embedded ICC profile.
```
svgeesus, line 15: Should perhaps explicitly note that this is exactly the same as [iCCP](https://w3c.github.io/PNG-spec/#11iCCP) except the profile name is UTF-8 instead of Latin-1. Analogous to [tEXt](https://w3c.github.io/PNG-spec/#11tEXt) vs. [iTXt](https://w3c.github.io/PNG-spec/#11iTXt)

svgeesus, line 15: Suggest, analogous to [iTXt](https://w3c.github.io/PNG-spec/#11iTXt)

"... are not permitted. The profile name shall not contain a zero byte (null character). The only ..."

svgeesus, line 15: I would like to understand the restriction to Output or Display profiles.

For example, [CSS `@color-profile`](https://drafts.csswg.org/css-color-4/#at-profile) says:

>  Input, Display, Output, or ColorSpace ICC profile. (Abstract, DeviceLink, and NamedColor ICC Profiles must not be used).

svgeesus:
> 
> 
> Can this made as a PR against https://github.com/w3c/ColorWeb-CG/blob/master/png_cicp_chunk_proposal.md ?

Why? They are entirely different chunk proposals?

svgeesus:
Oh - iCCN would need to be added to the list in [6.1 Colour types and values](https://w3c.github.io/PNG-spec/#6Colour-values), too.

> Greyscale samples represent luminance if the transfer curve is indicated (by gAMA, sRGB, or iCCP) or device-dependent greyscale if not. RGB samples represent calibrated colour information if the colour space is indicated (by gAMA and cHRM, or sRGB, or iCCP) or uncalibrated device-dependent colour if not

iCCN_for_PNG.md
```diff
@@ -0,0 +1,26 @@
...
+
+If the iCCN chunk is present, the image samples conform to the colour space represented by the embedded ICC profile. The colour space of the ICC profile shall be an RGB (`RGB `) colour space for colour images (PNG colour types 2, 3, and 6), or a greyscale (`GRAY`) colour space for greyscale images (PNG colour types 0 and 4). A PNG encoder that writes the iCCN chunk is encouraged to also write `gAMA` and `cHRM` chunks that approximate the ICC profile, to provide compatibility with applications that do not use the iCCN chunk. When the iCCN chunk is present, PNG decoders that recognize it and are capable of colour management shall ignore any `gAMA`, `cHRM`, and `CiCP` chunks and use the iCCN chunk instead and interpret it according to [ICC] or [ICC-2001] as appropriate. PNG decoders that are used in an environment that is incapable of full-fledged colour management should use the gAMA and cHRM chunks if present.
```
ProgramMax, line 17: 'CiCP' -> 'cICP'

lrosenthol, line 17: Break the paragraph before "When the iCCN chunk..."

lrosenthol: Are you sure about that @ProgramMax ?  I checked the other spec when I was writing - but I will double check either way.

iCCN_for_PNG.md
lrosenthol: The intent is to exclude iccMAX since there are still outstanding security concerns that I would like to avoid.

iCCN_for_PNG.md
```diff
@@ -0,0 +1,26 @@
...
+
+A PNG datastream shall contain at most one embedded profile, whether specified explicitly with an `iCCP` or `iCCN` chunk or implicitly with an `sRGB` chunk.
```
ProgramMax, line 19: I think we want to specify it like iCCN takes precedence over iCCP. We can still say that an encoder should not use both.

This would give a standard behavior.

palemieux:
# 2021-03-31

Winning proposal: the iCCN chunk shall take precedence over the cICP chunk if the display/surface does not support the codepoint specified in cICP. (and provides examples).

iCCN_for_PNG.md
lrosenthol: Same rationale - to restrict the types of profiles that can be used and that make sense in this context.

PR #31, opened 2021-04-29
iCCP and iCCN specify different ICC versions

### LINKED ISSUE #30: iCCP chunk is not yet ICCv4
The HDR in PNG spec says the only difference between iCCP and iCCN chunks is Latin-1 vs. UTF-8.
This is not completely correct. iCCP is supposed to be ICCv2. Many decoders will parse ICCv4, but that isn't part of the spec of the iCCP chunk.

So an additional difference between iCCP and iCCN is ICCv2 vs. ICCv4.

ProgramMax:
The HDR in PNG proposal currently says the only difference between
the iCCP and iCCN chunks is Latin-1 vs. UTF-8 profile names.

But this is not completely correct. The PNG spec explicitly says
iCCP is ICCv2. Although many decoders treat iCCP as ICCv4, the
iCCN chunk would be explicitly ICCv4.

This commit clarifies this distinction.

Closes #30


svgeesus:
Yes, the inadvertent restriction to ICC v.2 in the 2003 PNG specification is a [known erratum](https://github.com/w3c/PNG-spec/issues/22)

svgeesus (approved):

PR #38, opened 2021-05-08
Clarify color space definitions

### LINKED ISSUE #37: Conversion to XYZD50 space in the definition of rec2100-pq
> To convert rec2100-pq to XYZD50 under relative colorimetric intent, perform the following steps:

Why is the `rec2100-pq` colorspace defined in terms of a conversion to XYZD50?

Rec. ITU-R BT.2100, Tables 2 and 4 already unambiguously defines a mapping between `rec2100-pq` and XYZ coordinates. In other words, one can map any `(R', G', B')` triplet to an `(X, Y, Z)` triplet expressed in absolute luminance (cd/m<sup>2</sup>).

Is the idea to make `rec2100-pq` fit in the [skia](https://skia.org/docs/user/color/) framework?

If so, it is not clear that skia is sufficient/appropriate for the definition of `rec2100-pq` since, as far as I can tell, the XYZD50 connection colorspace is not display-referred, i.e. the coordinates in the chromaticity in the XYZD50 connection colorspace are in arbitrary units and not in absolute luminance.

palemieux:
The definitions of the color spaces are separated from the specification of the conversion between them.

Since the XYZD50 color system was introduced merely for mathematical expediency, it is replaced with the BT.2100 color system. This removes the need for a white point conversion since sRGB, HLG and PQ share the same D65 white point. 

Closes #37

### DIFF
--- a/hdr_html_canvas_element.md
+++ b/hdr_html_canvas_element.md
@@ -81,7 +81,7 @@ Add a new `CanvasStorageFormat` enum to allow for higher bit storage formats.
   enum CanvasStorageFormat {
     'unorm-8',
     'unorm-10-10-10-2',
-    'float-16", 
+    'float-16', 
   }
 ```
 
@@ -101,7 +101,9 @@ WebGPU's ``GPUSwapChainDescriptor`` can allow for specifying higher bit depth fo
 
 ### Color spaces
 
-Update `PredefinedColorSpace` to include the following new color spaces.
+#### General
+
+Update `PredefinedColorSpace` to include the following color spaces.
 
 ```idl
   partial enum PredefinedColorSpace {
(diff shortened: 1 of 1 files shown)


hdr_html_canvas_element.md
```diff
@@ -112,77 +114,100 @@ Update `PredefinedColorSpace` to include the following new color spaces.
...
+
+The component signals are mapped to red, green and blue tristimulus values according to the PQ system system specified in Rec. ITU-R BT.2100.
+
 ### Conversion between color spaces
 
-All `PredefinedColorSpace` are defined by how they are converted to XYZD50 under relative colorimetric intent.
+#### General
 
-Converting from XYZD50 to a `PredefinedColorSpace` is done by performing the inverse of the conversion to XYZD50.
+This section specifies the color space conversion when compositing element with color space A onto a canvas element with color space B.
```
ccameron-chromium, line 156: This sentence needs some work. While the intent is clear, it doesn't use the language of the HTML spec.
* The operation being discussed is [drawing](https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawimage) a [CanvasImageSource](https://html.spec.whatwg.org/multipage/canvas.html#canvasimagesource) in a canvas rendering context.
  * The image source has some color space, and the canvas rendering context has a color space.
  * If they are different, then the image source [must be converted](https://html.spec.whatwg.org/multipage/canvas.html#colour-spaces-and-colour-correction) to the canvas rendering context's before drawing.
* The term "compositing" refers to how things are displayed on the output device (monitor)
  * The general rule is that drawing an image source to a canvas and then compositing that canvas should be directly compositing the image source.
  * This rule cannot be obeyed for HDR content (because it depends on the final display, which is not known for offscreen canvases, and is not knowable due to fingerprinting)
* So what we're discussing is "what is that conversion that is done"
  * For the CSS predefined color spaces (that we want to glom on to here), the HTML specification specifies the conversion [here](https://www.w3.org/TR/css-color-4/#predefined-to-lab) 
  * That definition goes through XYZD50 (and then Lab)

If you want an off the shelf replacement for the above sentence, I'd go with this:

-- 

There are several places in the HTML specification where a color space conversion is required (e.g, when [drawing images to a canvas](https://html.spec.whatwg.org/multipage/canvas.html#colour-spaces-and-colour-correction), [retrieving image data from a canvas](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-getimagedata), among others being added). There exists a standard conversion that is applied to all data in these situations, namely, a conversion using relative colorimetric intent.

This operation is defined for [CSS Predefined Color Spaces](https://www.w3.org/TR/css-color-4/#predefined) in the HTML specification [here](https://www.w3.org/TR/css-color-4/#predefined-to-lab). For each color space, the conversion is parameterized by a transfer function that converts pixel values to a linear space, and a matrix that converts values in that linear space to values in CIE XYZ.

In this section we define this conversion for the new predefined color spaces.

hdr_html_canvas_element.md
```diff
@@ -112,77 +114,100 @@ Update `PredefinedColorSpace` to include the following new color spaces.
...
 
-Color values in XYZD50 may assume any real value (including values less than zero and greater than one).
+These conversions are expressed using a connection color space with the system colorimetry specified in Rec. ITU-R BT.2100:
```
ccameron-chromium, line 158: The HTML spec [uses XYZD50](https://www.w3.org/TR/css-color-4/#predefined-to-lab), so I'd just stick with that pattern.

At a minimum the connection color space should be one of the [CSS Predefined Color Spaces](https://www.w3.org/TR/css-color-4/#predefined) (or one of the color spaces we plan to add to that set), so that we don't have to re-define it here.

hdr_html_canvas_element.md
```diff
@@ -112,77 +114,100 @@ Update `PredefinedColorSpace` to include the following new color spaces.
...
-* Apply the matrix transformation to convert the [sRGB white point](https://www.w3.org/TR/css-color-4/#valdef-color-srgb) to the D50 white point.
+_Note:_ The domain of this transformation function is all real values. Its domain is not restricted to the unit interval [0, 1].
 
-Note that the domain of this transformation function is all real values. Its domain is not restricted to the unit interval [0, 1].
 Also see note in the Issues section at the bottom about whether this space should be distinct from the existing `'srgb'` space.
 
 #### `rec2100-hlg`
 
-To convert `rec2100-hlg` to XYZD50 under relative colorimetric intent, perform the following steps:
+* Transfer function: HLG Reference OETF specified at Rec. ITU-R BT.2100
```
ccameron-chromium, line 194: I think this should read the "inverse of the HLG Reference OETF".

(we should probably just write out the math)

hdr_html_canvas_element.md
```diff
@@ -112,77 +114,100 @@ Update `PredefinedColorSpace` to include the following new color spaces.
...
-If this is followed, then converting from  `rec2100-hlg` to any SDR color space will not result in any luminance clipping.
-This is a desirable property.
+* Connection matrix: Identity
+
+_Note:_ Converting from `rec2100-hlg` to any SDR color space will not result in clipping.
 
 #### `rec2100-pq`
 
-To convert `rec2100-pq` to XYZD50 under relative colorimetric intent, perform the following steps:
+* Transfer function: `EOTF<sup>-1</sup>[F<sub>D</sub>/300]` where `EOTF<sup>-1</sup>` is the inverse of the Reference PQ EOTF specified at Rec. ITU-R BT.2100.
```
ccameron-chromium, line 204: If 300 is our final point if disagreement, we'll be in a very good place :D

ccameron-chromium (commented):
A few suggestion to do before merging

hdr_html_canvas_element.md
palemieux: If the connection space is used merely for mathematical expediency, why use a more complicated mathematical expression and a white point which is not relevant for HDR color systems?

hdr_html_canvas_element.md
palemieux: I think the gain will need to be parameterized as done in [TTML2](https://www.w3.org/TR/ttml2/#style-attribute-luminanceGain). I plan to file a separate issue.

hdr_html_canvas_element.md
palemieux: The first step of the conversion is _apply the inverse transfer function of color space A_, i.e. linearize the color components. _HLG Reference OETF_ takes linear components and generates a non-linear components, so the _transfer function_ is _HLG Reference OETF_, right?

> (we should probably just write out the math)

Yes, it would be good to add as an informative section.

Issue #51, opened 2021-06-08
transfer function for extended-sRGB doesn't handle E' less than -0.04045


michaeldsmith:
negative values input to pow(value, exponent) function are not well defined when exponent is not an integer in current definition (exponent = 2.4 is not an integer)
```
   E = | E' / 12.92, if abs(E') ≤ 0.04045
       | ((E' + 0.055) / 1.055)^2.4, otherwise

       with E' ∈ ℝ 
```

Do you want symmetry around 0 but with negative sign for negative input, or do you want the linear segment to just keep going linearly forever into negative output for negative input?  

The if/else below gives symmetry around 0 with negative output for negative input, I think this is what you really want, right?

```
if N < -0.04045
  L = -1.0 * (((-1.0*N) + 0.055) / 1.055)^2.4;
elseif (N >= -0.04045 && N <= 0.04045)
  L  = N / 12.92;
else
  L = ((N + 0.055) / 1.055)^2.4;
end
```

![image](https://user-images.githubusercontent.com/37905569/121102071-e892b680-c7b1-11eb-8f53-d1f2fe0099a7.png)

![image](https://user-images.githubusercontent.com/37905569/121102091-f3e5e200-c7b1-11eb-86c9-2e5b26af13c2.png)

svgeesus:
> Do you want symmetry around 0 but with negative sign for negative input, or do you want the linear segment to just keep going linearly forever into negative output for negative input?

Most implementations mirror around 0. 

For example [CSS Color 4 defines it like this](https://www.w3.org/TR/css-color-4/#valdef-color-srgb):

```js
let sign = c < 0? -1 : 1;
let abs = Math.abs(c);

if (abs < 0.04045) {
  cl = c / 12.92;
}
else {
  cl = sign * (Math.pow((abs + 0.055) / 1.055, 2.4));
}
```

c is the gamma-encoded red, green or blue component. cl is the corresponding linear-light component.

svgeesus:
Note that separating out the sign and then adding it back is important because `pow` is not defined for negative values and will return `NaN`

ccameron-chromium:
Rolled in with https://github.com/w3c/ColorWeb-CG/pull/new/fix-abs

ccameron-chromium:
Thanks -- I've updated this to write the function as Javascript and to fix the missing sign issue in https://github.com/w3c/ColorWeb-CG/pull/70

PR #52, opened 2021-06-08
fix extended-srgb TF to handle negative values

michaeldsmith:
addresses issue #51

### DIFF
--- a/hdr_html_canvas_element.md
+++ b/hdr_html_canvas_element.md
@@ -153,7 +153,7 @@ The component signals are mapped to red, green and blue tristimulus values accor
 * Green primary chromaticity: `(0.300, 0.600)`
 * Blue primary chromaticity: `(0.150, 0.060)`
 * White point chromaticity: `(0.3127, 0.3290)`
-* Transfer function: `E = E', with E' ∈ ℝ` where `E'` is the non-linear colour value and `E` is the linear colour value
+* Transfer function: `L = N, with N ∈ ℝ` where `N` is the non-linear colour value and `L` is the linear colour value
 
 
 _NOTE:_ The color primary chromaticities, white point chromaticity and transfer function are those of the scRGB system specified at ISO/IEC 61966-2-2. Other aspects of ISO/IEC 61966-2-2 do not apply.
--- a/hdr_html_canvas_element.md
+++ b/hdr_html_canvas_element.md
@@ -125,13 +125,23 @@ The component signals are mapped to red, green and blue tristimulus values accor
 * Transfer function:
 
 ```
-   E = | E' / 12.92, if abs(E') ≤ 0.04045
-       | ((E' + 0.055) / 1.055)^2.4, otherwise
+decoding equation (non-linear to linear):
 
-       with E' ∈ ℝ
+if N < -0.04045
+  L = -1.0 * (((-1.0*N) + 0.055) / 1.055)^2.4;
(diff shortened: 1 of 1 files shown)


hdr_html_canvas_element.md
```diff
@@ -125,13 +125,23 @@ The component signals are mapped to red, green and blue tristimulus values accor
 * Transfer function:
 
 ```
-   E = | E' / 12.92, if abs(E') ≤ 0.04045
-       | ((E' + 0.055) / 1.055)^2.4, otherwise
```
ccameron-chromium, line 129: I'd prefer it to be

```
   E = | E' / 12.92, if abs(E') ≤ 0.04045
       | sign(E') * ((abs(E') + 0.055) / 1.055)^2.4, otherwise
```

That emphasizes the point symmetry better, in my opinion.

hdr_html_canvas_element.md
```diff
@@ -125,13 +125,23 @@ The component signals are mapped to red, green and blue tristimulus values accor
...
+end
 
-       with E' ∈ ℝ
 ```
 
-where `E'` is the non-linear colour value and `E` is the linear colour value
+where `N` is the non-linear colour value and `L` is the linear colour value
+
+![extended-srgb-decoding-function](https://user-images.githubusercontent.com/37905569/121105535-25ae7700-c7b9-11eb-92fd-a90e68cac7f3.png)
+![extended-srgb-decoding-function-linear-section](https://user-images.githubusercontent.com/37905569/121105539-27783a80-c7b9-11eb-8024-a8db2c110eb0.png)
```
ccameron-chromium, line 143: I don't feel that adding a figure emphasizing that the central section of `extended-srgb` is linear clarifies this proposal, so I would prefer to leave it out.

ccameron-chromium, line 143: I don't feel that adding a figure emphasizing that the central section of `extended-srgb` is linear clarifies this proposal, so I would prefer to leave it out.

hdr_html_canvas_element.md
```diff
@@ -125,13 +125,23 @@ The component signals are mapped to red, green and blue tristimulus values accor
...
+  L = ((N + 0.055) / 1.055)^2.4;
+end
 
-       with E' ∈ ℝ
 ```
 
-where `E'` is the non-linear colour value and `E` is the linear colour value
+where `N` is the non-linear colour value and `L` is the linear colour value
+
+![extended-srgb-decoding-function](https://user-images.githubusercontent.com/37905569/121105535-25ae7700-c7b9-11eb-92fd-a90e68cac7f3.png)
```
ccameron-chromium, line 142: IIUC embedded images should be uploaded into this repo (similar to what's done in [PR 61](https://github.com/w3c/ColorWeb-CG/pull/61)).

If we are to include an illustration of transfer functions, I'd prefer to have a single figure with all transfer functions included on it, something to the effect of [this](https://colab.research.google.com/drive/1UJvajR9mh7u5lWWzHMIyncQFsiIZykRm?usp=sharing). I'm not convinced about how helpful the illustration would be.

ccameron-chromium (changes requested):
Thanks for noting that the equation for transfer functions was incorrect. I'd prefer to do a minimal change here, and to leave out any plots of the transfer functions (at least for now).

svgeesus:
Now handled by https://github.com/w3c/ColorWeb-CG/pull/70

Issue #49, opened 2021-06-07
Add encoding transfer function equation to colorspace definitions


palemieux:

palemieux:
Assigned to @simontWork

PR #53, opened 2021-06-08
Add encoding transfer function equation to colorspace definitions

michaeldsmith:
addresses #49

### DIFF
--- a/hdr_html_canvas_element.md
+++ b/hdr_html_canvas_element.md
@@ -135,12 +135,23 @@ else
   L = ((N + 0.055) / 1.055)^2.4;
 end
 
+encoding equation (linear to non-linear):
+
+if L < -0.0031308
+  N = -1.055 * (-1.0 * L) ^ (1.0/2.4) + 0.055;
+elseif (L >= -0.0031308 && L <= 0.0031308)
+  N  = L * 12.92;
+else
+  N = 1.055 * L ^ (1.0/2.4) - 0.055;
+end
 ```
 
 where `N` is the non-linear colour value and `L` is the linear colour value
 
 ![extended-srgb-decoding-function](https://user-images.githubusercontent.com/37905569/121105535-25ae7700-c7b9-11eb-92fd-a90e68cac7f3.png)
 ![extended-srgb-decoding-function-linear-section](https://user-images.githubusercontent.com/37905569/121105539-27783a80-c7b9-11eb-8024-a8db2c110eb0.png)
+![extended-srgb-encoding-function](https://user-images.githubusercontent.com/37905569/121106265-89856f80-c7ba-11eb-9263-a2d634f4a047.png)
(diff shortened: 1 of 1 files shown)


hdr_html_canvas_element.md
```diff
@@ -125,13 +125,33 @@ The component signals are mapped to red, green and blue tristimulus values accor
 * Transfer function:
 
 ```
-   E = | E' / 12.92, if abs(E') ≤ 0.04045
-       | ((E' + 0.055) / 1.055)^2.4, otherwise
-
-       with E' ∈ ℝ
+decoding equation (non-linear to linear):
```
ccameron-chromium, line 128: I would prefer to only have a single function defined, and to indicate that the conversion from space X to Y is the inverse function of the conversion of space Y to X.  This behavior (that transformations are always inverses) is a bit in contrast to how some things are done in the broadcast world, and I'd like to be very clear about this behavior, and adding the definition of the inverse (at least for now) might take away from that emphasis.

hdr_html_canvas_element.md
```diff
@@ -125,13 +125,33 @@ The component signals are mapped to red, green and blue tristimulus values accor
...
+  N  = L * 12.92;
+else
+  N = 1.055 * L ^ (1.0/2.4) - 0.055;
+end
 ```
 
-where `E'` is the non-linear colour value and `E` is the linear colour value
+where `N` is the non-linear colour value and `L` is the linear colour value
+
+![extended-srgb-decoding-function](https://user-images.githubusercontent.com/37905569/121105535-25ae7700-c7b9-11eb-92fd-a90e68cac7f3.png)
```
ccameron-chromium, line 151: (same remarks about plots as in [PR 52](https://github.com/w3c/ColorWeb-CG/pull/52))

ccameron-chromium (commented):
Sorry for the delay in getting to this. I'd prefer to leave it out for now.

hdr_html_canvas_element.md
palemieux: Per 2021-07-19: make the decoding equation normative and the encoding equation informative/example.

hdr_html_canvas_element.md
palemieux: 2021-07-19: either have graphs for all functions or none.

hdr_html_canvas_element.md
```diff
@@ -125,13 +125,33 @@ The component signals are mapped to red, green and blue tristimulus values accor
 * Transfer function:
 
 ```
-   E = | E' / 12.92, if abs(E') ≤ 0.04045
-       | ((E' + 0.055) / 1.055)^2.4, otherwise
-
-       with E' ∈ ℝ
+decoding equation (non-linear to linear):
+
+if N < -0.04045
```
palemieux, line 130: Per 2021-07-19: clarify the range and domain of N and L.

PR #56, opened 2021-07-12
Spell out color space chunk priority

### LINKED ISSUE #54: `cICP` and `sRGB`
I'm mainly thinking about allowed, disallowed chunk combinations, and priority of chunks where there is a potential conflict.

From [Adding support for HDR imagery to the PNG format](https://github.com/w3c/ColorWeb-CG/blob/master/hdr-in-png-requirements.md)

> A PNG datastream shall contain at most one embedded profile, whether specified explicitly with an iCCP or iCCN chunk or implicitly with an sRGB chunk.

This raises an interesting distinction between a color profile (which provides numerical data about a color space, but does not name it) and a color space description (which names, but relies on external knowledge for the numerical data).

For example, given two ICC profiles, which declare the same (or very similar) primary chromaticities and whitepoint, and very similar transfer function (but there are many ways that could be expressed - as a LUT, and a parametric function, etc) are these **the same colorspace**

The answer has to be "no" or at least "we don't know"; while for labeling schemes, given two labels, they unambiguously either are or are not the same.

`cICP` is clearly labeling, and `iCCP` and `iCCN` are clearly profiles. I had always considered `sRGB` to be labeling and thus, similar to `cICP` except that it conveys only one colorspace option instead of many. The current text considers `sRGB` to be an implicit color profile. I think it does so primarily to constrain the combinations of chunks. According to that logic isn't `cICP` also an implicit profile?

I seems to me that if there are two PNG files, one has `sRGB` and the other has `cICP` with values 1/13/0/1 then they are both unambiguously labeled as the sRGB color space. The proposal says you can have both, but then the `sRGB` must be ignored.

I don't think `sRGB` should be described as an implicit profile, now that we have the concept of color space labelling.

ProgramMax:
This commit spells out color space chunk priority. With that clear
priority, it allows multiple color space chunks.

Closes #54

### DIFF
--- a/hdr-in-png-requirements.md
+++ b/hdr-in-png-requirements.md
@@ -18,7 +18,7 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
 
 ## Strawman approach
 
-A PNG may contain multiple chunks with color space information. A PNG viewer should use the highest priority color space chunk, ignoring the others. The priorities are from highest to lowest:
+A PNG may contain multiple chunks with color space information. A PNG viewer should use the highest priority color space chunk that it can honor, ignoring the others. The priorities are from highest to lowest:
 
 * `cICP`
 * `iCCN`
@@ -74,7 +74,7 @@ If the `iCCN` chunk is present, the image samples conform to the colour space re
 
 ### Decoder
 
-The `iCCN` chunk should be interpreted according to [ICC] or [ICC-2010] as appropriate. PNG decoders that are used in an environment that is incapable of full-fledged colour management shall use the `gAMA` and `cHRM` chunks if present.
+The `iCCN` chunk should be interpreted according to [ICC] or [ICC-2010] as appropriate. 
 
 ## References
 
--- a/hdr-in-png-requirements.md
+++ b/hdr-in-png-requirements.md
@@ -48,10 +48,6 @@ The `cICP` chunk SHALL come before `IDAT` chunk.
 
 #### Structure
 
-The four-byte chunk type field contains the decimal values
-
-`105 67 67 80`
-
 The `iCCN` chunk contains:
 
 | Profile name |  1-79 bytes (character string)
--- a/hdr-in-png-requirements.md
+++ b/hdr-in-png-requirements.md
@@ -18,6 +18,15 @@ An existing W3C group note, [BT2100-in-PNG]  specifies an approach which is limi
 
 ## Strawman approach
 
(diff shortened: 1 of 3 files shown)


hdr-in-png-requirements.md
```diff
@@ -72,11 +74,7 @@ If the `iCCN` chunk is present, the image samples conform to the colour space re
 
 ### Decoder
 
-If the image contains a `cICP` chunk and will be rendered to a display or surface that supports `cICP`, then the PNG decoder shall ignore any `gAMA`, `cHRM`, `iCCP`, and `iCCN` chunks and use the `cICP` chunk instead. Otherwise, when a `iCCN` chunk is present, PNG decoders that recognize it and are capable of colour management shall ignore any `gAMA`, `cHRM`, and `cICP` chunks and use the `iCCN` chunk instead and interpret it according to [ICC] or [ICC-2010] as appropriate. PNG decoders that are used in an environment that is incapable of full-fledged colour management shall use the `gAMA` and `cHRM` chunks if present.
-
-#### Codestream
-
-A PNG datastream shall contain at most one embedded profile, whether specified explicitly with an `iCCP` or `iCCN` chunk or implicitly with an `sRGB` chunk.
+The `iCCN` chunk should be interpreted according to [ICC] or [ICC-2010] as appropriate. PNG decoders that are used in an environment that is incapable of full-fledged colour management shall use the `gAMA` and `cHRM` chunks if present.
```
palemieux, line 77: > PNG decoders that are used in an environment that is incapable of full-fledged colour management shall use the `gAMA` and `cHRM` chunks if present.

What is the objective of this clause? A decoder can also use `cICP` if present.

palemieux (changes requested):

hdr-in-png-requirements.md
ProgramMax: I think the intent is that "incapable of full-fledged color management" means a true color profile cannot be upheld.

That said, maybe it would be best to spell out that a PNG decoder should go down the priority list until it finds a chunk that it can honor. Thoughts?

palemieux: > That said, maybe it would be best to spell out that a PNG decoder should go down the priority list until it finds a chunk that it can honor. Thoughts?

+1

ProgramMax:
I managed to get into rebase hell on this pull request.
I made a clean pull request to replace it: https://github.com/w3c/ColorWeb-CG/pull/60

PR #74, opened 2021-12-06
Correct priority order of PNG chunks.

ProgramMax:
The existing HDR in PNG requirements doc lists the 'sRGB' chunk as
lowest priority.

However, the PNG spec [1] says that the 'sRGB' chunk has a higher priority
than 'gAMA' and 'cHRM', as it is short-hand for an exact set of 'gAMA'
and 'cHRM' values.

This commit corrects the priority order of the PNG chunks.

[1] https://www.w3.org/TR/2003/REC-PNG-20031110/#11sRGB


svgeesus (approved):
Yes, this is correct. Thanks for the fix!

PR #76, opened 2022-03-28
Update HLG and PQ constants

ccameron-chromium:
Specify that the signal of 0.75 HLG and ref luminance 203 PQ should match SDR white.


palemieux (approved):
Reviewed at 2022-03-28 meeting

PR #77, opened 2022-06-13
Clarify cICP fallthrough

ProgramMax:
If the cICP specifies a code point which the decoder knows cannot be
used by the display / surface, the cICP chunk should be ignored. This
commit adds that wording.

### DIFF
--- a/hdr-in-png-requirements.md
+++ b/hdr-in-png-requirements.md
@@ -30,7 +30,7 @@ A PNG may contain multiple chunks with color space information. A PNG viewer sho
 * `iCCP`
 * `gAMA` & `cHRM`
 
-Note: If the PNG decoder knows the display / surface cannot use the cICP code point, it should ignore the cICP chunk and continue down the priority list.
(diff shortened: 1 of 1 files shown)


hdr-in-png-requirements.md
```diff
@@ -30,6 +30,8 @@ A PNG may contain multiple chunks with color space information. A PNG viewer sho
 * `iCCP`
 * `gAMA` & `cHRM`
 
+Note: If the PNG decoder knows the display / surface cannot use the cICP code point, it should ignore the cICP chunk and continue down the priority list.
```
svgeesus, line 33:
```suggestion
If the PNG decoder knows the display / surface cannot use the cICP code point, it should ignore the cICP chunk and continue down the priority list.
```

Notes are informative, not normative, so can't be used to express requirements.

svgeesus (changes requested):

svgeesus:
@ProgramMax could you make the requested change, please?

svgeesus (approved):

PR #89, opened 2022-11-30
CG Charter

### LINKED ISSUE #88: Charter
Does this CG have a charter? I wasn't able to find one.

palemieux:
The following were used as a starting point:
* [Kick-off message to the mailing list](https://lists.w3.org/Archives/Public/public-colorweb/2017Jan/0001.html)
* [README.md](https://github.com/w3c/ColorWeb-CG/blob/master/README.md)

Closes #88

### DIFF
--- a/charter.html
+++ b/charter.html
@@ -73,10 +73,6 @@
         <li>Consistency and predictability of display</li>
       </ul>
     </p>
-    <h3 id="out-of-scope">
-      Out of Scope
-    </h3>
-    <p><i>n/a</i></p>
     <h2 id="deliverables">
       Deliverables
     </h2>
@@ -110,9 +106,9 @@
       <ul>
         <li><a href="https://www.w3.org/groups/wg/htmlwg">W3C HTML WG</a></li>
         <li><a href="https://www.w3.org/groups/wg/png">W3C PNG WG</a></li>
-        <li><a href="https://www.w3.org/2011/webtv/wiki/Main_Page">W3C Media and Entertainment Interest Group</a></li>
+        <li><a href="https://www.w3.org/groups/ig/me">W3C Media and Entertainment Interest Group</a></li>
         <li><a href="https://www.w3.org/groups/wg/gpu">W3C GPU for the Web Working Group</a></li>
-        <li><a href="https://www.w3.org/media-wg/">W3C Media Working Group</a></li>
+        <li><a href="https://www.w3.org/groups/wg/media">W3C Media Working Group</a></li>
         <li><a href="https://github.com/whatwg/html">WHATWG HTML Workstream</a></li>
         <li><a href="https://github.com/AOMediaCodec">Alliance for Open Media</a></li>
         <li><a href="https://github.com/KhronosGroup/WebGL">Khronos WebGL</a></li>
--- a/charter.html
+++ b/charter.html
@@ -115,6 +115,7 @@
         <li><a href="https://www.w3.org/media-wg/">W3C Media Working Group</a></li>
         <li><a href="https://github.com/whatwg/html">WHATWG HTML Workstream</a></li>
         <li><a href="https://github.com/AOMediaCodec">Alliance for Open Media</a></li>
+        <li><a href="https://github.com/KhronosGroup/WebGL">Khronos WebGL</a></li>
       </ul>
     </p>
     <h2 id="process">
--- a/charter.html
+++ b/charter.html
@@ -113,6 +113,8 @@
         <li><a href="https://www.w3.org/2011/webtv/wiki/Main_Page">W3C Media and Entertainment Interest Group</a></li>
         <li><a href="https://www.w3.org/groups/wg/gpu">W3C GPU for the Web Working Group</a></li>
         <li><a href="https://www.w3.org/media-wg/">W3C Media Working Group</a></li>
+        <li><a href="https://github.com/whatwg/html">WHATWG HTML Workstream</a></li>
+        <li><a href="https://github.com/AOMediaCodec">Alliance for Open Media</a></li>
       </ul>
     </p>
     <h2 id="process">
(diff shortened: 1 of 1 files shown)


annevk:
Liaisons should probably list the WHATWG HTML Workstream, GPU for the Web Working Group, and Khronos? And maybe AOM?

chrisn:
I suggest also adding Media WG for WebCodecs.

charter.html
```diff
@@ -0,0 +1,285 @@
...
+        peak whites greater than paper white)</li>
+        <li>Color matching between media (raster images, video, vector graphics)
+        and text/styling content (styled by CSS)</li>
+        <li>Consistency and predictability of display</li>
+      </ul>
+    </p>
+    <h3 id="out-of-scope">
+      Out of Scope
+    </h3>
+    <p><i>n/a</i></p>
```
annevk, line 79: Can we drop this section? This makes it seem like everything could potentially be in scope.

hober: Instead of dropping it, I'd rather it contain a concise list of things that are definitely out of scope, e.g.

* anything unrelated to color
* native platform color things unrelated to the web

or some such.

palemieux: The scope is _The works addresses the representation and processing of color information as applied to the Web_, which excludes topics unrelated to the intersection of web and color.

Do you have very specific examples of topics that would be out-of-scope.

palemieux:
I have added W3C GPU for the Web WG and W3C Media WG.

annevk:
@palemieux in practice I would expect the Color CG to directly interact with the WHATWG HTML Workstream as per the W3C/WHATWG Memorandum.

I suggested AOM for the various media formats that also include color representation and Khronos for its WebGL work, which also relies on color representation.

palemieux:
> in practice I would expect the Color CG to directly interact with the WHATWG HTML Workstream as per the W3C/WHATWG Memorandum.

Ok. What is the mode of contact/interaction with WHATWG HTML Workstream?

> I suggested AOM for the various media formats that also include color representation and Khronos for its WebGL work, which also relies on color representation.

Ok. We need a point of contact in both orgs.

annevk:
@palemieux you essentially file an issue on the GitHub repository (whatwg/html). I think the same works for AOM and WebGL.

palemieux:
> you essentially file an issue on the GitHub repository (whatwg/html)

This one?

https://github.com/whatwg/html

> I think the same works for AOM

Ok. I do not think we can add a liaison until we know how to contact them.

annevk:
Yeah. For AOM you'd file an issue on the relevant repo as well: https://github.com/AOMediaCodec. I've contacted them that way in the past successfully!

annevk:
@palemieux apologies for assuming familiarity before. I think we were talking past each other a bit. Khronos would be https://github.com/KhronosGroup/WebGL. Thanks for making the changes!

palemieux:
@annevk Added liaison Khronos WebGL

palemieux:
@annevk Hopefully you will be able to join the CG to help with these liaisons.

palemieux:
>  If we're adding AOM, should we also consider adding ITU-T SG16 and SMPTE? Whether we actually need contact with those groups I guess depends on the extent we need to coordinate with them in practice.

I am reluctant to adding liaisons for which there are no volunteers that commit to managing them.

svgeesus:
The "Out of Scope" section in charters is for things that **look** like they might be in scope, but are not. For example because another group or org is already doing that. As an example the Web Fonts WG specifically excludes changes to OFF/OpenType, which are instead done by liaison to SC29.

If there is nothing in that category (as here) then the section is simply deleted.

charter.html
svgeesus: The "Out of Scope" section in charters is for things that look like they might be in scope, but are not. For example because another group or org is already doing that. As an example the Web Fonts WG specifically excludes changes to OFF/OpenType, which are instead done by liaison to SC29.

If there is nothing in that category (as here) then the section is simply deleted.

charter.html
```diff
@@ -0,0 +1,290 @@
...
+    </p>
+    <h2 id="liaisons">
+      Dependencies or Liaisons
+    </h2>
+    <p>
+      The group expects to liaise with the following groups:
+      <ul>
+        <li><a href="https://www.w3.org/groups/wg/htmlwg">W3C HTML WG</a></li>
+        <li><a href="https://www.w3.org/groups/wg/png">W3C PNG WG</a></li>
+        <li><a href="https://www.w3.org/2011/webtv/wiki/Main_Page">W3C Media and Entertainment Interest Group</a></li>
```
svgeesus, line 113: This one should be the same format as the other links to W3C groups:

https://www.w3.org/groups/ig/me

charter.html
```diff
@@ -0,0 +1,290 @@
...
+        <li><a href="https://www.w3.org/groups/wg/gpu">W3C GPU for the Web Working Group</a></li>
+        <li><a href="https://www.w3.org/media-wg/">W3C Media Working Group</a></li>
```
svgeesus, line 115: Same here

https://www.w3.org/groups/wg/media

charter.html
```diff
@@ -0,0 +1,290 @@
...
+      The <a href="https://www.w3.org/Consortium/cepc/">W3C Code of
+      Ethics and Professional Conduct</a> applies to participation in
+      this group.
+    </p>
+
+    <h2 id="worklimit">
+      Work Limited to Charter Scope
+    </h2>
+    <p>
+      The group will not publish Specifications on topics other than those
```
svgeesus, line 154: This means that every time we add a new specification we need to update the charter. But that is okay.

svgeesus (approved):

palemieux:
The consensus during the 2022-12-05 telecon is to keep the list of liaison organizations as-is for now.

PR #95, opened 2023-02-24
Minimal HDR Canvas proposal

palemieux:
Refactor the HDR Canvas proposal to a minimal baseline focused on today's HDR imagery and practice without precluding future optimizations and extensions.

Highlights:

- add support for HLG and PQ color spaces, including explicit recommendations for:
  - tone-mapping from HLG and PQ to sRGB
  - mapping from sRGB to HLG and PQ
  - mapping between HLG and PQ
- add support 16-bit floats bitmaps


hdr_html_canvas_element.md
```diff
@@ -71,679 +84,276 @@ Add a new color volume structure for use in several places.
...
 
-This section describes the default compositing behavior for canvas element.
-This is the behavior that happens if the `CanvasColorMetadata` is not set, or if is called specifying `'default'` as the mode.
+## Annex A: Color space conversions
 
-In this mode, the `HTMLCanvasElement` will be composited exactly as an `<img>` or `<video>` element with a source in the canvas' color space would be composited.
+### Background
 
-This means that if the canvas' color space is `'rec2100-hlg'` or `'rec2100-pq'`, then the canvas will be composited using high dynamic range, where available.
+In general, application should avoid conversions between color spaces and
```
chrisn, line 241: "application" -> "applications"

hdr_html_canvas_element.md
```diff
@@ -71,679 +84,276 @@ Add a new color volume structure for use in several places.
...
+maintain imagery in its original color space: conversions between color spaces
+are not necessarily reversible and do not necessarily result in the same image
+appearance. In particular, conversion of an HDR image to an SDR will result in a
```
chrisn, line 244: "to an SDR" -> "to SDR"

hdr_html_canvas_element.md
```diff
@@ -71,679 +84,276 @@ Add a new color volume structure for use in several places.
...
+signification loss of information and an SDR image that is different from the
```
chrisn, line 245: "signification" -> "significant"

hdr_html_canvas_element.md
```diff
@@ -71,679 +84,276 @@ Add a new color volume structure for use in several places.
...
+SDR image that would have been mastered from the same source material. From that
+perspective, converting from HDR to SDR imagery is similar to converting RGBA
+images to 16-color pallette images.
```
chrisn, line 248: "pallette" -> "palette"

chrisn (commented):
Just reviewing today, so I added a few more comments

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,313 @@
...
-    required double greenPrimaryY;
-    required double bluePrimaryX;
-    required double bluePrimaryY;
-    required double whitePointX;
-    required double whitePointY;
-  }
-```
+* drawing images retrieved from a file whose format is not supported by the
+  `img` or `video` elements
+* collages of images, both HDR and SDR, for a variety of sources
```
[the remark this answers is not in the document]
palemieux: I removed "variety of sources" since it was vague and unnecessary: it does not really matter where the pixels come from.

The idea is that the compositing happens in whatever colorspace the canvas is in, i.e. the incoming pixels are first converted to the Canvas' colorspace and then composited.

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,313 @@
...
+* drawing of UI elements that accompany an HDR video presentation
```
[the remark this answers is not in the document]
palemieux: "accompany" means "related". I agree that UI was too specific. I have also clarified that the ultimate objective is for the visual elements to match the video presentation.

hdr_html_canvas_element.md
```diff
@@ -126,9 +131,18 @@ bitmap](https://html.spec.whatwg.org/multipage/canvas.html#output-bitmap) of a
...
-`colorType = "unorm8"` corresponds to HTML Canvas as it exists today.
+When `colorType = "unorm8"`, the non-linear component signals {R', G', B'} are
+quantized using full range quantization, i.e. they are multiplied by 255 and
+rounded to the nearest integer. For example, R' = 0.5 is represented by the
+integer value 128.
+
+When `colorType = "float16"`, the non-linear component signals {R', G', B'} are
+not quantized, i.e. R' = 0.5 is represented by the floating point value 0.5.
+
+_NOTE: `colorType = "unorm8"` corresponds to HTML Canvas as it exists today ans
```
chrisn, line 142: Typo: ans -> and

hdr_html_canvas_element.md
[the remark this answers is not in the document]
palemieux: @chrisn Ok?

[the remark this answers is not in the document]
chrisn: Looks good, thanks

hdr_html_canvas_element.md
```diff
@@ -121,101 +122,36 @@ For WebGPU this would be:
...
+(using relative colorimetric intent), all of its color components are within the `[0, 1]` interval. When displaying content produced
+by a rendering context with `CanvasHighDynamicRangeMode` set to `'default'`, all colors that are within the screen's default range
+must be displayed without any clamping or roll-offs (neither in chrominance nor luminance).
 
-A color is said to be within a screen's extended range if, when that color is converted to the screen's native linear color space (using relative colorimetric intent), all of its color components are within the `[0, highDynamicRangeHeadroom]` interval.
-When displaying content produced by a rendering context with `CanvasHighDynamicRangeMode` set to `'extended'`, all colors that are within the screen's extended range must be displayed without any clamping or roll-offs (neither in chrominance nor luminance).
+A color is said to be within a screen's *extended* range if, when that color is converted to the screen's native linear color space
+(using relative colorimetric intent), all of its color components are within the `[0, highDynamicRangeHeadroom]` interval. When
```
svgeesus: That definition carefully excludes out of gamut colors, which would have some negative values, while leaving in-gamut HDR values. Nice.

hdr_html_canvas_element.md
```diff
@@ -121,101 +122,36 @@ For WebGPU this would be:
...
+displaying content produced by a rendering context with `CanvasHighDynamicRangeMode` set to `'extended'`, all colors that are within
+the screen's extended range must be displayed without any clamping or roll-offs (neither in chrominance nor luminance).
```
svgeesus: That precludes any soft rolloff or tonemapping. Is that intentional?

hdr_html_canvas_element.md
```diff
@@ -266,167 +197,85 @@ The component signals are mapped to red, green and blue tristimulus values accor
...
-* Linear light scaling: 1.0
-
-#### `srgb`
+### Color space conversions
 
-* Transfer function: See [srgb-linear](#srgb-linear)
+#### Background
 
-* Connection matrix: matrix to convert the [sRGB primaries](https://www.w3.org/TR/css-color-4/#valdef-color-srgb) to the primary colours specified in Rec. ITU-R BT.2100
+In general, application should avoid conversions between color spaces and maintain imagery in its original color space: conversions
```
svgeesus: But canvas always converts on input from the specified colorspace to that of the canvas.

hdr_html_canvas_element.md
```diff
@@ -266,167 +197,85 @@ The component signals are mapped to red, green and blue tristimulus values accor
...
+canvas](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-getimagedata), among others being added). The conversions
+between predefined SDR color spaces are defined at <https://www.w3.org/TR/css-color-4/>, and this proposal similarly defines
+conversions for HDR color spaces.
 
-* Connection matrix: Identity
+These conversions fall into two broad categories:
 
-#### `rec2100-pq`
+* conversion between HDR color spaces
+* conversion between an HDR and an SDR color space (tone mapping)
```
svgeesus, line 242: Tone mapping can also be the result of the HDR headroom of the display device being less than that of the mastering display, no?

hdr_html_canvas_element.md
```diff
@@ -266,167 +197,85 @@ The component signals are mapped to red, green and blue tristimulus values accor
...
- 2. Convert from `extended-srgb` color space to `rec2100-hlg` color space
- 3. Scale pixel values - See Note 1
- 4. Apply HLG Inverse EOTF to convert to HLG from Pseudo-Display Light with Lw = 302 cd/m2 - See Note 1
-  * apply HLG Inverse OOTF
-  * apply HLG OETF - See Note 2
 
-_Note 1:_ As `rec2100-hlg` is a relative format, the brightness of the virtual monitor used for mathematical transforms can be chosen to be any level. In this transform it is chosen to be 302 cd/m2 so that the brightness of diffuse white matches the diffuse white of sRGB (80 cd/m2). Using the extended range gamma formula in footnote 2 of BT.2100, this also sets the HLG Inverse OOTF to be unity. The value 0.265 is calculated by taking the inverse OETF of 0.75, the `rec2100-hlg` diffuse white level.
-
-_Note 2:_ See section 5.3 in ITU-R BT.2408-4 relating to negative transfer functions in format conversions.
+1. Apply HLG EOTF to convert the non-linear `rec2100-hlg` Signal to linear Pseudo-Display Light with Lw = 302 cd/m2 - See Note 1
```
svgeesus: Typo: 302 should be 203

hdr_html_canvas_element.md
```diff
@@ -266,167 +197,85 @@ The component signals are mapped to red, green and blue tristimulus values accor
...
+_Input:_ Full-range non-linear floating-point `srgb-linear` pixel with black at 0.0 and diffuse white at 1.0. Values may exist outside the range 0.0 to 1.0.
 
 _Output:_ Full-range non-linear floating-point `rec2100-hlg` pixel with black at 0.0 and diffuse white at 0.75. Values may exist outside the range 0.0 to 1.0.
 
 _Process:_
-1. Convert from `extended-linear-srgb` color space to `rec2100-hlg` color space
+
+1. Convert from `extended-srgb-linear` color to `rec2100-hlg` color
 2. Scale pixel values - See Note 1
 3. Apply HLG Inverse EOTF to convert to HLG from Pseudo-Display Light with Lw = 302 cd/m2 - See Note 1
```
svgeesus: Again should be 203

hdr_html_canvas_element.md
```diff
@@ -435,114 +284,118 @@ function convertExtendedLinearSRGBtoREC2100HLG(r, g, b) {
...
   return [r5, g5, b5];
 }
 ```
 
-##### Conversion from extended-sRGB to PQ
-
-_Input:_ Full-range non-linear floating-point `extended-srgb` pixel with black at 0.0 and diffuse white at 1.0. Values may exist outside the range 0.0 to 1.0.
-
-_Output:_ Full-range non-linear floating-point `rec2100-pq` pixel with black at 0.0 and diffuse white at ???. Values may exist outside the range 0.0 to 1.0.
+_Note 1:_ As `rec2100-hlg` is a relative format, the brightness of the virtual monitor used for mathematical transforms can be chosen to be any level. In this transform it is chosen to be 302 cd/m2 so that the brightness of diffuse white matches the diffuse white of sRGB (80 cd/m2). Using the extended range gamma formula in footnote 2 of BT.2100, this also sets the HLG Inverse OOTF to be unity. The value 0.265 is calculated by taking the inverse OETF of 0.75, the `rec2100-hlg` diffuse white level.
```
svgeesus: 203

hdr_html_canvas_element.md
```diff
@@ -435,114 +284,118 @@ function convertExtendedLinearSRGBtoREC2100HLG(r, g, b) {
...
+
+_Note:_ The factor `k` is selected such that a display luminance of 203 cd/m<sup>2</sup> results in a linear color value of 1 in `srgb-linear`.
 
-_Input:_ Full-range non-linear floating-point `rec2100-hlg` pixel with black at 0.0 and diffuse white at 0.75. Values may exist outside the range 0.0 to 1.0.
+##### `srgb-linear` to `rec2100-pq`
+
+_Input:_ Full-range non-linear floating-point `extended-srgb` pixel with black at 0.0 and diffuse white at 1.0. Values may exist outside the range 0.0 to 1.0.
 
-_Output:_ Full-range non-linear floating-point `extended-srgb` pixel with black at 0.0 and diffuse white at 1.0. Values may exist outside the range 0.0 to 1.0.
+_Output:_ Full-range non-linear floating-point `rec2100-pq` pixel with black at 0.0 and diffuse white at ???. Values may exist outside the range 0.0 to 1.0.
```
svgeesus: What to say for ??? 
PQ diffuse white is at whatever the creative intent places it at, so maybe instead say peak small-area white (10,000 cd/m2) at 1.0?

svgeesus (commented):
A few minor comments

hdr_html_canvas_element.md
palemieux: Yes. The text here however applies narrowly to the color space conversion that are covered by this explainer, which does not deal with what happens at the display.

hdr_html_canvas_element.md
palemieux: I think you are looking at an old version of the document.

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,336 @@
...
-    }
-    else {
-      return = sign * (Math.pow((abs + 0.055) / 1.055, 2.4));
-    }
-  }
-```
+## Add HDR rendering behavior and HDR metadata to `CanvasRenderingContext2DSettings`
 
-#### srgb-linear
+Add a new CanvasColorMetadata dictionary:
```
RafaelCintron, line 172: How will we help developers provide good values for the dataype and colorspace of the canvas?  I originally thought the spec would have recommendations based on the metadata but the metadata is only available after the canvas is created.  

Do we expect developers to first make an sacrificial canvas, see what metadata is returned, and then make the "real" canvas afterwards?  Seems like the metadata should be available independent of an individual canvas instance. 

In general, I think the spec should provide use cases for the metadata such as how the metadata guides what `DOMString` values to pass for 2D canvas colors. Without use cases, this is going to draw the ire of those who oppose any fingerprinting bits be exposed to web developers for informational purposes. I, personally, think we can make the case for having these but we need to be more clear why it is needed in the spec.

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,336 @@
...
-      return 0;
-    } else if (x <= 0.5) {
-      return x * x / (3 * k);
-    } else if (x <= 1) {
-      return (Math.exp((v - c) / a) + b) / (12 *  k);
-    } else {
-      return 1 / scale;
-    }
+```idl
+// SMPTE ST 2086 color volume metadata.
```
RafaelCintron, line 196: Please define what "SMPTE ST 2086" stands for somewhere in the spec.

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,336 @@
...
+dictionary CanvasSmpteSt2086Metadata {
```
RafaelCintron, line 197: Does "SmpteSt2086" need to be part of the enum name? Is there some other metadata we plan to add in the future that needs its own dictionary?  I think calling this `CanvasMetadata` should suffice.

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,336 @@
...
-    required double whitePointX;
-    required double whitePointY;
-  }
-```
+* drawing images retrieved from a file whose format is not supported by the
+  `img` or `video` elements
+* collage of images, both HDR and SDR
+* drawing of visual elements that are related to an HDR video presentation, with
+  the expectation that the visual elements match the look of the HDR video
+  presentation.
```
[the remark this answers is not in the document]
palemieux: Is the following clearer?

_drawing of visual elements that are related to an HDR presentation in a `<video>` element, with the expectation that the visual elements match the look of the `<video>` element presentation_

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,336 @@
...
-  const b1 = simpleTransform(b, systemGamma);
-  const [r2, g2, b2] = matrixXYZtoSRGB(matrixBT2020toXYZ(r1, g1, b1));
-  const r3 = simpleInverseTransform(r2, systemGamma);      
-  const g3 = simpleInverseTransform(g2, systemGamma); 
-  const b3 = simpleInverseTransform(b2, systemGamma); 
-  const [r4, g4, b4] = limitTosRGBGamut(r3, g3, b3);
-  return [r4, g4, b4];
-}
+```idl
+  enum CanvasColorType {
```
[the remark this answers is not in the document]
palemieux: I agree. Done.

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,336 @@
...
+    "unorm8",
+    "float16",
```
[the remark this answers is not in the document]
palemieux: > Is this a hedge against float16 not getting approved by the JS standards body? 

I think it is saying that it is not an essential part of this spec, and that either would work -- at least initially. I do not like however the idea of a placeholder, so listing the two seemed the least terrible approach. What is your preference?

hdr_html_canvas_element.md
```diff
@@ -1,749 +1,336 @@
...
+`ImageData`.
 
 ```idl
-  partial enum PredefinedColorSpace {
-    'srgb-linear',
-    'rec2100-hlg',
-    'rec2100-pq',
-  }
+  partial dictionary ImageDataSettings {
+    ImageDataColorType colorType = "Uint8ClampedArray";
```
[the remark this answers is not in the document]
palemieux: It should be "unorm". Thanks.

hdr_html_canvas_element.md
palemieux: Yes, we definitely need more information -- similar to what is provided in the PNG 3ED spec: https://w3c.github.io/PNG-spec/#mDCv-chunk and https://w3c.github.io/PNG-spec/#cLLi-chunk

There are two basic use cases:

* the application receives the ST 2086 metadata, e.g. from metadata within the image file or from an API call. In this case, the application simply copies the information as in the `CanvasSmpteSt2086Metadata` structure
* the application generates the ST 2086 metadata, in which case, there are best practices we should be able to point to

palemieux: See added prose

michaeldsmith:
The ST2094-10 Annex B method mentioned in the section "rec2100-pq to srgb" 

image

needs parameters min/avg/max values.  

The example code at https://www.sandflow.com/public/tone-mapping/index.html performs image analysis to determine the min/avg/max values. This is a great way to obtain the min/avg/max parameters for a single image.  

I think that it should be suggested that the good results will come from image analysis.  

If image analysis is not performed, then the min/avg/max values need to be provided as parameters to 2094-10 Annex B another way.  Default min and max values could come from ST2086 mastering display metadata min luminance and max luminance, default max value could come from MaxCLL metadata.  The average will depend on if the scene is bright, average or dark.  Using a default average value between 5-15 nits could be a good starting place.

ccameron-chromium:
This removes the features related to rendering content in an SDR-relative, application-tone-mapped canvas (in particular, the `extended` mode, and the [`ScreenDetails`](https://www.w3.org/TR/window-management/#api-screendetails-interface) parts.

Splitting these into two proposals is probably the right thing to do -- there is too much misunderstanding between the two worlds that both call themselves HDR.

I would also suggest separating the `float16` support out from this proposal. It already has its own proposal [here](https://github.com/w3c/ColorWeb-CG/blob/main/canvas_float.md) and lengthy-and-stalled discussion [here](https://github.com/whatwg/html/issues/8708).

A remaining core part of this proposal (with this PR applied) is a standardization of how to render PQ and HLG content on phone/tablet/laptop/desktop operating systems.

These operating systems parameterize their displays by two parameters: HDR headroom and color gamut (this encompasses of Windows, iOS, macOS, Android, and ChromeOS, including the built-into-a-TV variants). None of the existing standards address this problem (they have no concept of a time-varying HDR headroom, which is fundamental to all operating systems' implementations of HDR).

There is also [this issue](https://github.com/whatwg/html/issues/9112) about standardizing how PQ and HLG images are to be rendered as HTMLImageElement.

palemieux:
> I would also suggest separating the float16 support out from this proposal.

I do not think this is separable since HDR cannot exist in an 8-bit world, as described in the introduction.

>  there is too much misunderstanding between the two worlds that both call themselves HDR.

I do not see a misunderstanding.

HLG and PQ have been iterated for years and are used by all HDR TV and movie content. The web has become an important source of TV and movie consumption. It makes sense for the web platform to support them.

This does not preclude/conflict with other approaches, as far as I know.

ccameron-chromium:
> > I would also suggest separating the float16 support out from this proposal.
> 
> I do not think this is separable since HDR cannot exist in an 8-bit world, as described in the introduction.

Changes to `PredefinedColorSpace` for HLG and PQ is valuable on its own, because `PredefinedColorSpace` is used by 2D canvas, ImageData, WebGL, and WebGPU.

There are some issues to be worked out for 2D canvas and ImageData with respect to float16 support, but WebGPU is already shipping support for it, and WebGL has [a proposal](https://github.com/KhronosGroup/WebGL/pull/3222) that doesn't have any issues that I know of.

I wouldn't want us to block HLG and PQ support in WebGPU because of unrelated 2D canvas issues (which may take a long time to resolve).

I also don't think we should explicitly disallow 8-bit HDR (it's not unreasonable for HLG -- see [these 8-bit JPEG HLG images](https://ccameron-chromium.github.io/hdr-jpeg/index.html)).

palemieux:
> I also don't think we should explicitly disallow 8-bit HDR

Have you read the article at https://ieeexplore.ieee.org/document/7514362 ? It lays out the fact that 8-bit is not even sufficient for sRGB at low luminance.

P.S.: The images show up HDR in my Chrome browser! :)

palemieux:
> I wouldn't want us to block HLG and PQ support in WebGPU because of unrelated 2D canvas issues (which may take a long time to resolve).

I argue we should use HDR as a way to drive the resolution of these issues.

kenrussell:
We could almost replace references to `Float16Array` with `Float32Array` and not only would no web developer notice, but also the references could be changed back to `Float16Array` once that has been shipped in browsers with no application impact. Since reads from `Float16Array` will immediately convert to 64-bit float (JavaScript's native number representation) and writes to `Float16Array` silently truncate from that representation, it will behave very similarly to how `Float32Array` does today.

However - since this might be contentious, getting this in to have a spec browsers can start implementing would be preferable in my opinion.

svgeesus:
> I also don't think we should explicitly disallow 8-bit HDR

I disagree. Don't give people a footgun, and instead, use this to drive the higher-bit-depth canvas upgrade.

PR #105, opened 2023-07-25
Do not change default return type for getImageData

ccameron-chromium:


chrisn:
This change would leave the proposal without a way to get the image data for a float16 backed canvas. Is the idea to propose adding a `getHDRImageData()` or similar?

ccameron-chromium:
> This change would leave the proposal without a way to get the image data for a float16 backed canvas. Is the idea to propose adding a `getHDRImageData()` or similar?

No.

The function [getImageData](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData) takes an `ImageDataSettings` parameter. In this proposal we have added `colorType` to `ImageDataSettings`.

The following code will always return sRGB float32 data, regardless of the color space or pixel format of the canvas itself.
```
  myData = context.getImageData(x, y, w, h, {colorSpace:"srgb", colorType:"float32"})
```

chrisn:
I see, thanks!

palemieux:
I think this PR is in line with https://github.com/w3c/ColorWeb-CG/blob/main/hdr_html_canvas_element.md#extend-imagedatasettings-to-support-higher-bit-depths where the default value for `dataType`  is `"unorm8"`.

@ccameron-chromium Can you confirm?

ccameron-chromium:
> I think this PR is in line with https://github.com/w3c/ColorWeb-CG/blob/main/hdr_html_canvas_element.md#extend-imagedatasettings-to-support-higher-bit-depths where the default value for `dataType` is `"unorm8"`.
> 
> @ccameron-chromium Can you confirm?

Yes, that is correct, these result in the same behavior.

(There is a slight difference in that one defines how things behave when this is not specified, which could be because no ImageDataOptions is specified, or because the ImageDataOptions does not specify dataType).

palemieux (approved):

PR #106, opened 2023-09-05
Add interface to query the display color volume and linear display light encoding 

### LINKED ISSUE #36: Add HDR metadata to rec2100-pq canvases
Given the objective:

> To display PQ encoded HDR images and video in a canvas such that PQ images drawn in the canvas will appear exactly the same as they would if displayed via an  or <video> tag.

It should be possible for the application to (optionally) associate the following information with a `rec2100-pq` canvas:

## Mastering Display Color Volume Metadata as specified in SMPTE ST 2086

* Mastering Display Primaries: list of 3 (x, y) chromaticity coordinates. Each chromaticity coordinate consists of two uint16 elements, in order, the normalized x and y chromaticity coordinates of the color primary in units of 0.00002

EXAMPLE: Color primaries of the P3D65 system specified in SMPTE ST 2113: `{ 34000, 16000 } { 13250, 34500 } { 7500, 3000 }`

* Mastering Display White Point Chromaticity: (x, y) chromaticity coordinate. Each chromaticity coordinate consists of two uint16 elements, in order, the normalized x and y chromaticity coordinates of the color primary in units of 0.00002

EXAMPLE: Illuminant D65 specified in SMPTE RP 177: `{ 15635, 16450 }`

* Mastering Display Maximum Luminance: uint32 value in units of 0.0001 cd/m<sup>2</sup>

EXAMPLE: Reference white in Recommendation ITU-R BT.2035: `1000000`

* Mastering Display Maximum Luminance: uint32 value in units of 0.0001 cd/m<sup>2</sup>
  
EXAMPLE: Reference black in Recommendation ITU-R BT.2035: `100`

## MaxCLL and MaxFALL as specified in Annex P.1 and P.2 of CTA 861-G

* Maximum Content Light Level: uint16 in units of 1 cd/m<sup>2</sup>

* Maximum Frame-average Light Level: uint16 in units of 1 cd/m<sup>2</sup>

### LINKED ISSUE #39: Luminance gain factor required when compositing SDR bitmaps onto a PQ canvas
The sRGB color space specifies an absolute reference white point luminance of 80 cd∙m<sup>-2</sup>. As a result, colors specified in the sRGB  space cannot be reliably composited with display-referred high-dynamic range (HDR) images, such as those that use the perceptual quantizer electro-optical transfer function (PQ EOTF) specified in Rec. ITU-R BT.2100.

Specifically,
* sRGB colors may appear too dim if they remain unscaled and the average luminance of the scene exceeds 80 cd∙m<sup>-2</sup>; and
* sRGB colors may appear too bright if they are uniformly scaled to an arbitrary large luminance, such as 10,000 cd∙m<sup>-2</sup>.

Suggest following the same approach as used in [TTML2](https://www.w3.org/TR/ttml2/#style-attribute-luminanceGain)

### LINKED ISSUE #42: An image editor needs to know the capabilities of the mastering display
When color grading images, the user need to know if the image they see is the undistorted image or if the appearance is distorted by the display bounds. 

Non-web image editing apps (such as Photoshop) use the knowledge of the display gamut to 
- highlight which pixels are outside the display gamut.
- map such colors to similar in gamut colors without a huge hue shift.

The same capabilities are required in a web based image editing app.

The application needs this info:
- color primaries and white point
- max luminance or headroom above sRGB white
- min luminance

This is found in the SMPTE ST 2086 parameter set.
Add link here to ST 2086

### LINKED ISSUE #50: Provide concrete examples of color space transformations
* sRGB to `rec2100-hlg` (see https://www.w3.org/TR/ttml2/#hdr-compositing as a starting point)
* `extended-linear-srgb` to `rec2100-hlg
`

### LINKED ISSUE #92: Improve the name of the hdrToSdrRatio/highDynamicRangeHeadroom property
`hdrToSdrRatio`/`highDynamicRangeHeadroom` is intended to specify the maximum linear color component value after which clipping may occur, with the assumption that a value of 1.0 represents max component value in SDR.

Suggest changing the name to reflect the semantics above, e.g. `maxLinearComponentValue`.

### LINKED ISSUE #99: minLuminance and maxLuminance should explicitly refer to the BT.2100 viewing environment
From @ppaalanen

> I think this should explicitly refer to the BT.2100 viewing environment. When this content is being displayed in some other viewing environment, the cd/m² may not be emitted literally even on capable equipment.

https://github.com/w3c/ColorWeb-CG/blob/8eac62f15e5b41940636beb031f6214b61e933af/hdr_html_canvas_element.md?plain=1#L225

_Originally posted by @ppaalanen in https://github.com/w3c/ColorWeb-CG/pull/98#discussion_r1212970448_

### LINKED ISSUE #100: Should guidance on gamut mapping be provided?
The strawman provides examples of algorithms designed to render high dynamic range images to a lower dynamic range display -- the focus of the strawman is on high-dynamic range imagery.

These example algorithms do not tackle the mapping of wide gamut images to a display with a narrower gamut, e.g. from Rec, 2020 to Rec. 709 color gamut.

Should such examples be provided?

See also https://gitlab.freedesktop.org/pq/color-and-hdr/-/issues/18 .

_Originally posted by @ppaalanen in https://github.com/w3c/ColorWeb-CG/issues/98#issuecomment-1571844917_

### LINKED ISSUE #102: Provide MinTML/MaxTML metadata about the viewing environment
From what I can tell, the current `minimumLuminance` and `maximumLuminance` fields are metadata from the application *to* the renderer about the content, and the rendering environment applies the mapping to the display environment. As noted in #100, the standard currently has no input on how this tone and gamut mapping is applied to the display environment. On all current displays, the available theoretical range of gamut and luminance is much larger than the actual space the display itself can support.

Gamut and tone mapping is a lossy transform with no one correct answer; it's feasible for applications might want to do tone mapping themselves to make sure that their colors reproduce as accurately as possible. This is the approach [recommended by the HGIG](https://www.hgig.org/doc/ForBetterHDRGaming.pdf), who proposed making three metadata fields available to content: MinToneMapLuminance, MaxToneMapLuminance, and MaxFullFrameToneMapLuminance. Should we consider exposing these fields to applications that wish to provide their own tonemapping?

Historically, it's worth noting that the content-provided metadata approach was tried in SMPTE ST 2108; however it was effectively deprecated by all known platforms because the lack of standardized tonemapper across display environments made accurate color reproduction basically impossible.

### LINKED ISSUE #103: Add section on HDR to HDR tone-mapping
The strawman should explicitly mention HDR to HDR tone-mapping, as opposed to HDR to SDR tone-mapping. In particular, Annex 5 of Rep. ITU-R BT.2408-5 is also intended to work for HDR -> HDR tone-mapping.

See https://github.com/w3c/ColorWeb-CG/issues/102#issuecomment-1612466185

### LINKED ISSUE #109: rec2100-pq-to-srgb algorithm does not handle colors outside of 709 color primaries
https://github.com/w3c/ColorWeb-CG/blob/main/hdr_html_canvas_element.md#rec2100-pq-to-srgb

From: https://github.com/w3c/ColorWeb-CG/pull/106#issuecomment-1768982518

### LINKED ISSUE #110: rec2100-pq to srgb function uses Rec. 709 gamma (2.4) instead of sRGB gamma (2.2)
```
From Lars:
Why is PQ to srgb using gamma 2.4 while HLG uses 2.2? 2.2 is usually the better fit for sRGB. But maybe you want 709 brightness levels?
```

_Originally posted by @palemieux in https://github.com/w3c/ColorWeb-CG/issues/106#issuecomment-1768982979_

palemieux:
Closes #110 
Closes #109 
Closes #103 
Closes #102
Closes #100
Closes #99
Closes #92
Closes #50 
Closes #42 
Closes #39
Closes #36


hdr_html_canvas_element.md
```diff
@@ -283,6 +280,57 @@ the Content Light Level Information chunk can provide more accurate information
 than the maximum luminance parameter of the Mastering Display Color Volume
 chunk.
 
+As specified below, the platform does not generally apply color volume mapping
+if the color volume of the image is smaller than that of the display.
```
ppaalanen, line 284: On one hand this makes sense, you want to present the content as-is when possible.

OTOH, a faithful absolute colorimetric presentation does not produce then intended appearance if the viewing environment differs.

What are "platform" and "display" in this context, do they include the OS window system?

I can think of three different viewing environments in the pipeline:
- the content colorimetry has been authored for a mastering viewing environment A
- a monitor expects a video signal to be tuned for a reference viewing environment B
- the monitor adjusts the image, based on e.g. the end user brightness and contrast knobs, to look as intended in the actual viewing environment C

BT.2100 seems to have a common reference viewing environment for both PQ and HLG, so if both content and video signal follow either of them, a simple format conversion suffices. Otherwise, something more should probably happen between A and B. That you have covered by implying the viewing environment via `PredefinedColorSpace`, I guess.

From B to C may still be a mapping.

It's also possible that a monitor would not do image adjustment, which means the video signal would need to be tuned for viewing environment C directly. I'm not quire sure how PQ mode monitors/TVs behave. Maybe that depends on the vendor and model even.

All in all, I would be a bit wary of promising "generally no mapping".

ppaalanen (commented):
My comments here are more like food for thought rather than concrete review comments or even suggesting changes. Maybe you are aware of all of it, or maybe something should be considered, I'm not sure. Hope it helps.

palemieux:
Per 2023-09-06 meeting: add support for enumerated values of the color volume,

hdr_html_canvas_element.md
palemieux: I see multiple use cases:
1. the application provides an HLG or PQ image as-is and the platform performs additional adjustments if needed (the platform might actually perform no adjustments if, for example, the image is full screen and the display supports HLG or PQ natively)
2. (a) the platform provides to the application the nominal min/max luminance of the display, i.e. headroom, (b) the application performs tone-mapping, and (c) the platform and the display may or may not perform additional adjustments (ambient light color, user preference, etc.)
3. (a) the platform provides the min/max luminance of the display, ambient light level and color, etc., (b) the application performs all adjustments (c) the platform performs no adjustments (the display might still perform adjustments unbeknownst to the platform)

I would think that (1) should be supported no matter what.

ppaalanen: I don't think cases 1, 2 and 3 are really separate that much. The common thing is that the image source (application) describes the conditions of the image it has, and then the operating and window system do the best they can with it. If the application got some information in order to do a better image mapping, then the used information is part of the described image conditions. Of course, I can only speak for Wayland.

We're not really disagreeing here. All I'm saying is that "the platform does not generally apply color volume mapping
if the color volume of the image is smaller than that of the display" is a bit much to promise, and also a bit vague. If people start depending on it, you have to strike the word "generally". Otherwise the promise is moot, because it might not always hold.

We won't have that promise on Wayland. The only thing we would promise is that if you tag your image with the output's image description (color space, etc.) exactly, then  the image would be sent out in the video signal without adjustments, provided it is not blended with any other window and not a subject of compositor effects. (One reason for this is to allow applications do their own complete color management, another use case is display profiling.)

hdr_html_canvas_element.md
```diff
@@ -283,6 +280,57 @@ the Content Light Level Information chunk can provide more accurate information
...
+
+```idl
+partial interface Sceeen {
+  optional ColorVolumeInfo displayColorVolume;
+}
+```
+
+If present, the `displayColorVolume` attribute specifies the set of colors that
+the screen of the output device can reproduce without significant color volume
+mapping.
```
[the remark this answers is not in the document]
palemieux: Yes, I think `displayColorVolume` should be used only for high-performance applications that require precise and accurate knowledge of the display capabilities; for everything else, why not use `dynamic-range` and `color-gamut` media queries?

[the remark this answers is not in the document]
palemieux: ... there is also a much larger issue, which was brought up in the past: knowing the gamut and luminance range of a display is not really useful without also knowing the viewing environment (ambient light level, color, etc.). So it begs the question of how useful `displayColorVolume` is without also a means of parameterizing the viewing environment.

Specifying a complete output color system, e.g. Rec. ITU-T BT.2100 PQ, addresses this since such color system include a viewing environment. Note that the signaled viewing environment (dark surround in the case of  Rec. ITU-T BT.2100 PQ) might not match the current viewing environment (broad daylight), leaving the task to adapting the image to the platform/display.

Do we need to go into this level of detail for out initial pass at an HDR Canvas API?

hdr_html_canvas_element.md
palemieux: > All I'm saying is that "the platform does not generally apply color volume mapping
if the color volume of the image is smaller than that of the display" is a bit much to promise, and also a bit vague. If people start depending on it, you have to strike the word "generally". Otherwise the promise is moot, because it might not always hold.

I agree that the wording is vague.

AFAIK the idea is that, to the extent that the platform provide information on display capabilities, the platform should be encouraged to not alter colors that fit within the display's capabilities -- unless instructed otherwise by the user, etc.

Is your point is that this is not testable and that there are too many scenarios?

> The only thing we would promise is that if you tag your image with the output's image description (color space, etc.) exactly, then the image would be sent out in the video signal without adjustments, provided it is not blended with any other window and not a subject of compositor effects. 

Could this requirement (i.e. commit to a specific output color system and not modify images that match this color system) be reasonably imposed on the Web Platform in narrow scenarios, e.g., in full screen mode?

Maybe this is too narrow a use case to care about?

ppaalanen: > Is your point is that this is not testable and that there are too many scenarios?

I guess that depends on how strict you want to be. If you want to be lax, then maybe explain what the goal or intention is rather than trying to set rules. It is difficult to infer intention from rules, but a well communicated intention should get everyone on the same page.

Another thing is that color volume includes luminance. I'm not sure if this means absolute or relative luminance. I'm worried that this might be recommending presentation that is not adapted to the viewing conditions.

> Could this requirement (i.e. commit to a specific output color system and not modify images that match this color system) be reasonably imposed on the Web Platform in narrow scenarios, e.g., in full screen mode?

I'm not sure I understood the question, so pardon the wall of text.

Full screen mode is necessary, if we want to relay video HDR metadata straight to the display, because the metadata affects all of the screen. For Wayland desktop compositors that is annoying to implement, because changes in display driving mode or even HDR metadata (EOTF particularly) may take a while to sink in (ha!) or even blank the display for a moment.

Hence it is much more likely that a Wayland compositor chooses one mode (and static metadata set) for driving a display, and converts everything from applications to that. A Wayland compositor may be able to use the display controller hardware to do some or all of the conversion, so it may be possible to avoid GPU work in video playback.

This also means that full screen becomes irrelevant. A compositor converts all contents from all applications to the same, um, compositing color space that is tuned for each display. Even windowed video etc. should look fine and consistent.

If an application like a browser tags its window (or a video plane) with the output's image description, it only means that the Wayland compositor can skip most of the conversions and all automatic perceptual adjustments. Composition still happens in the compositing color space, but round-tripping to that is a no-op, and may even be skipped completely. But now the burden is on the application to follow the compositor's output image description, and there are only two good reasons to do that:
- the application is generating content (e.g. a game) and can choose an arbitrary image description to target for no or little extra cost, or
- the application wants to avoid the automatic conversions of a compositor for whatever reason.

For video playback performance using the output image description for video player's window content is detrimental, because it likely forces the video player to use a GPU rendering pass for color conversion, which means there cannot be any opportunity to off-load that conversion to the display controller hardware. Hence, one should pass video frames as-is from a hardware video decoder all the way to the window system, tagging them with the video's image description.

Display controller hardware color conversions are fixed-function and highly optimized for performance and low power consumption, which makes them preferable to the power hungry GPU when at all possible.

----

Referring to the diagram in https://gitlab.freedesktop.org/pq/color-and-hdr/-/merge_requests/35 , we often cannot and might not even care to control the emitted light directly, exactly because then we also need to take viewing conditions into account ourselves. As much as it pains me to leave the actual light undefined, it makes things much simpler if we can trust that whoever receives our standardized signal also takes care of adapting it appropriately.

This dilemma is  the core when trying to support both entertainment and ICC workflows in the same framework. Display-referred "do exactly as I say" vs. scene-referred and display-referred automatically adapted to the display and viewing conditions at hand. In Wayland it seems we have actually let go of even trying to guarantee the actual light at protocol level. If an end user wants that level of control for their applications, they are required to profile (measure) their monitors and use a particular workflow.

Uhh... am I even on topic here?

hdr_html_canvas_element.md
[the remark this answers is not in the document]
ppaalanen: Good question.

What if you just took a complete color system as a whole, like BT.2100 PQ and HLG? Leave room to have a custom color system as well, if it turns out that parameterising everything and targeting actual light becomes a use case.

That's what HDMI and DisplayPort essentially do.

hdr_html_canvas_element.md
```diff
@@ -285,25 +285,49 @@ if the color volume of the image is smaller than that of the display.
...
+}
+```
+
+If present,
+
+* `colorVolume` specifies, as defined above, the set of colors that the screen
+of the output device can reproduce without significant color volume mapping; and
+* `referenceWhiteLuminance` specifies the luminance of reference white as
+reproduced by the screen of the output device, and reflects viewing environment
+and user settings.
```
swick, line 309: From a whole-system kind-of view there are a few different ways to implement this...

1. The source targets a reference display and a reference viewing environment. The compositor works with the reference conditions. The sink itself is responsible for converting to the actual conditions, possibly with the help of a human changing settings on the sink.
2. The source targets a reference display and a reference viewing environment. The compositor is responsible for converting from the reference conditions to the actual conditions. The sink shows the signal as-is (ideally).
3. The source targets the actual conditions directly. The compositor does not change the signal. The sink shows the signal as-is.


Your sentence works just fine for all cases. In the first two, all the `ScreenColorInfo` values would be fixed all the time and the display/user will do the adjustment. In the second case the values will be dynamic.

While this probably shouldn't be part of the spec, we really should make sure that it's understood that those values do not necessarily imply real-world luminances and real-world display color volume.

Not making it clear has already caused a generation of desktop displays not having adjustable HDR modes.

palemieux: > While this probably shouldn't be part of the spec, we really should make sure that it's understood that those values do not necessarily imply real-world luminances and real-world display color volume.

I have tried writing such a sentence a couple of times, and have not found the right words. Maybe something as informal and short as you suggest above would work.

palemieux:
From Lars:

```
Typos?
CURRENT:

function simpleTransform(value, systemGamma) {

  if (value < 1.0) {

    return -1.0 * Math.pow(-1.0 * value, systemGamma);

  } else {

    return Math.pow(value, systemGamma);

  }

}

 

function simpleInverseTransform(value, systemGamma) {

  if (value < 1.0) {

    return -1.0 * Math.pow(-1.0 * value, 1.0 / systemGamma);

  } else {

    return Math.pow(value, 1.0 / systemGamma);

  }

}

 ```

**Comment by palemieux** [comment]
```
I belive pq to srgb needs a similar code snippet to map to sRGB gamut:
 

const r3 = simpleInverseTransform(r2, systemGamma);

  const g3 = simpleInverseTransform(g2, systemGamma);

  const b3 = simpleInverseTransform(b2, systemGamma);

  const [r4, g4, b4] = limitTosRGBGamut(r3, g3, b3);
```

**Comment by palemieux** [comment]
From Lars:
```
Why is PQ to srgb using gamma 2.4 while HLG uses 2.2? 2.2 is usually the better fit for sRGB. But maybe you want 709 brightness levels?
```

hdr_html_canvas_element.md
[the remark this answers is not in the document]
simontWork: As Pierre says, the viewing environment is built in to the signal for PQ and needs to be adapted, for HLG the equations for adaptation are given within the ITU document.

The main use case for this metadata is for when the user/operating system adjusts the diffuse white level.  You can't just scale the image as that may clip the highlights and will change the perception of the shadows and mid tones.  For example, if you have a 1000 nit monitor, HLG 75% signal level would normally appear at 203 nits.  If you increase the diffuse white to 350 and expect the HDR image or video to follow, you would need either a brighter display, or you need to tonemap the highlights to prevent clipping.  (I'm working on some example code for this)

simontWork (approved):
Agree with the code updates, need to add teonmapping at a later date.

ccameron-chromium (commented):
This looks like a good step forward.

In general I would prefer to avoid attempting to expose min/max/white luminance, and instead specify HDR headroom, but we can work that separately.

I'd also argue for using the conversion formulation in https://github.com/w3c/ColorWeb-CG/issues/111

hdr_html_canvas_element.md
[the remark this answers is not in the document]
ppaalanen: Right, but where can we find this `displayColorVolume` / `ScreenColorInfo` metadata?

A browser could expect to get it from the OS, but where would the OS get it from?

PR #125, opened 2024-06-20
Clarify expectactions for `toDataURL()` and `toBlob()`

### LINKED ISSUE #119: Is the color space preserved during toBlob() and toDataURL()
From https://github.com/w3ctag/design-reviews/issues/917#issuecomment-1964765613

The used color space is a property of CanvasRenderingContext2D, but HTMLCanvasElement provides methods for exporting the canvas contents as an image (toBlob(), toDataURL()). Is the color space preserved during these operations? (again this also applies to wide gamut)

palemieux:
Closes #119

### DIFF
--- a/hdr_html_canvas_element.md
+++ b/hdr_html_canvas_element.md
@@ -162,6 +162,10 @@ not quantized, i.e. R' = 0.5 is represented by the floating point value 0.5.
 _NOTE: `dataType = "unorm8"` corresponds to HTML Canvas as it exists today and
 should not be used to represent HDR signal, as detailed in the introduction._
 
+The `HTMLCanvasElement` methods `toDataURL()` and `toBlob()` should produce
+resources that best match the fidelity of the underlying HTML Canvas element
+(subject to the limitations of the implementation and the requested format).
+
 ## Extend `ImageDataSettings` to support higher bit depths
 
 Add to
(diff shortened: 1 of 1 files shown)


hdr_html_canvas_element.md
```diff
@@ -162,6 +162,10 @@ not quantized, i.e. R' = 0.5 is represented by the floating point value 0.5.
 _NOTE: `dataType = "unorm8"` corresponds to HTML Canvas as it exists today and
 should not be used to represent HDR signal, as detailed in the introduction._
 
+The `HTMLCanvasElement` methods `toDataURL()` and `toBlob()` should produce
+resources that best match the fidelity of the underlying HTML Canvas element
+(subject to the limitations of the implementation and the requested format).
```
svgeesus, line 167: That seems a bit vague, hand-wavy and hard to test.

It seems better to crisply state that the color space of the 2D context should be preserved in `toDataURL()` and `toBlob()`. (And yes, if the requested format can't do that then there is an issue, but all common formats can represent that information).

svgeesus (changes requested):
This doesn't really respond to the TAG issue as it is too vague (and I don't see a reason for the vagueness). Currently, a Canvas 2D context is either sRGB or Display P3 and that info is easy to persist (for example using CICP).

palemieux:
How can it be made less vague without providing a specific mapping to each format, which the Web Platform does not do today?>

svgeesus:
> How can it be made less vague 

By explicitly stating that the color space of the Canvas context should be preserved, rather than vaguely mentioning "fidelity"

> without providing a specific mapping to each format, which the Web Platform does not do today?

The mechanism by which each format does this need not be specified in detail. There are two common cross-format methods in use and my suggested text mentions both as examples.

svgeesus (approved):

PR #127, opened 2025-01-29
Add color-hdr() CSS proposal.

ccameron-chromium:


svgeesus:
@ccameron-chromium

ccameron-chromium:
Thanks for the ping. I'm linking to the CSS WG version of this now, rather than this explainer, so I'll close this without merge.

PR #137, opened 2026-02-12
Add updates for TAG review

ccameron-chromium:
Add updates for the TAG review (including security and privacy, alternates considered).

Also update the diagram to move CSS images and CSS into the "Potentially HDR content" block on the left, rather than having a separate block in the center for it. Add text to the "HTML element" block indicating that it isn't just for `img` and `video`, but also any other type, potentially pulling in things like background-image, etc.


svgeesus (approved):
LGTM

